PR #140, opened 2014-12-15
Fix bug where terminal button fails if spaces in Kitematic.app path

jmorganca:


rabbah:
Is this bug back?

bash: HD/Users/me/.docker/machine/machines/dev: No such file or directory
my path is /Volumes/Macintosh HD/Users/me/.docker/machine/machines/dev

PR #148, opened 2014-12-28
added helper for displaying tags as labels

schmunk42:
I played around a little bit with the Kitematic source. 
This is not a perfect solution, if you don't like the changes code-wise, please take it as a feature request.

I just thought, that KM could display the tag information in a more condensed way.
It may also be more suitable if not the helper extracts the tags, but some other function beforehand, but I didn't dive too deep into the code.

Anyway, here's how it would look like:

bildschirmfoto 2014-12-28 um 01 11 31

bildschirmfoto 2014-12-28 um 01 07 46

### DIFF
--- a/meteor/client/main.js
+++ b/meteor/client/main.js
@@ -197,6 +197,23 @@ Handlebars.registerHelper('displayTags', function (tags, delimiter) {
   }
 });
 
+Handlebars.registerHelper('displayTagsAsLabels', function (tags) {
+  if (tags) {
+    // extract name
+    name = tags[0].substr(0, tags[0].indexOf(':'));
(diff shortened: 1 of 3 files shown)


razmo:
:+1:

razmo:
Just a question how do you get this interface because when i install the version 0.5.7 (latest) i have this interface 

screen shot 2015-03-05 at 11 09 03

mchiang0610:
@razmo We did a redesign of Kitematic for v0.5.x

jeffdm:
Thanks for the contribution, @schmunk42! This feature's been added upstream as we've changed the foundation of the app from Meteor to React.js:

See https://github.com/kitematic/kitematic/blob/master/src/ImageCard.react.js

Thanks again.

PR #161, opened 2015-01-27
Add tag dropdown for creating containers with a specific image tag

jeffdm:
Also move container styles into container-modal.less


Elesant:
LGTM.

PR #198, opened 2015-02-23
Fix upgrading

jeffdm:
- Better wording for upgrading VirtualBox. Please review wording @Elesant @mchiang0610 
- Fixes a bug where if VirtualBox was already installed and a boot2docker VM was running the install would fail


Elesant:
LGTM!

PR #204, opened 2015-02-27
Use Docker Machine to create VMs

jeffdm:
- Use Machine instead of boot2docker to provision a VM
- Bump some packages
- Output zip files as -zip instead


Elesant:
Super awesome! LGTM!

PR #244, opened 2015-03-12
spell

jyu-rmn:


jeffdm:
Thank you! Unfortunately there's still a typo - my mistake from before.

It should be "A bug is reproducible and..."

PR #281, opened 2015-03-15
Altering the repo name to be a link that takes you to the Docker Hub page

allen-servedio:
Altered the repo name on the ImageCard to open the Docker Hub page for the repo when it is clicked on. I tried to follow the style that you all use for the tag link - though I am not sure that is how you want it to be styled (feel free to change to conform to the standards you are trying to set).

A more advanced version of this might be to open that page within another window.

### DIFF
--- a/src/ImageCard.react.js
+++ b/src/ImageCard.react.js
@@ -42,11 +42,11 @@ var ImageCard = React.createClass({
     $tagOverlay.fadeOut(300);
   },
   handleRepoClick: function () {
-    var $repoUri = 'https://registry.hub.docker.com/'
+    var $repoUri = 'https://registry.hub.docker.com/';
     if (this.props.image.is_official) {
-      $repoUri = $repoUri + "_/"
+      $repoUri = $repoUri + "_/";
     } else {
-      $repoUri = $repoUri + "u/"
+      $repoUri = $repoUri + "u/";
     }
     util.exec(['open', $repoUri + this.props.image.name]);
   },
@@ -67,7 +67,7 @@ var ImageCard = React.createClass({
       name = (
         <div>
           <div className="namespace official">{namespace}</div>
-          <OverlayTrigger placement="bottom" overlay={<Tooltip>{this.props.image.name}</Tooltip>}>
+          <OverlayTrigger placement="bottom" overlay={<Tooltip>View on DockerHub</Tooltip>}>
             <span className="repo" onClick={this.handleRepoClick}>{repo}</span>
           </OverlayTrigger>
         </div>
--- a/src/ImageCard.react.js
+++ b/src/ImageCard.react.js
@@ -5,6 +5,7 @@ var ContainerStore = require('./ContainerStore');
 var metrics = require('./Metrics');
 var OverlayTrigger = require('react-bootstrap').OverlayTrigger;
 var Tooltip = require('react-bootstrap').Tooltip;
+var util = require('./Util');
 
 var ImageCard = React.createClass({
   getInitialState: function () {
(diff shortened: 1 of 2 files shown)


DieterReuter:
Just tested it, works great. :+1:
But the graphical response distracts me a little, it overlaps the text way too much. Maybe we have an UX designer here, who could help to get the best visual effect.

@jeffdm: Is it possible to open such an external web link in-place in the apps view? Sometimes it's really annoying, when you'll get a new browser windows for each and every click. I just would like to see both ways side-by-side, to come to a better decision what's more convenient for the users.

allen-servedio:
Yep, I agree with you. It might be a good idea to form up a basic style guide or something so that contributions remain consistent with the over all UX experience.  I think having the page appear as a window in the app makes sense (the overviews are just not enough info for people to pick a container). I am also thinking of adding this to running containers for folks to be able to see what the container is based on.

Elesant:
Thank you so much for this! Definitely a great feature. Will figure out a design and style guide that makes sense. :+1:

allen-servedio:
Great and thanks. If you all are open to it, I will try to add things as I hit them. For example,  I may take a swag at pulling the dockerfile and let them view it right in the app. Developers like details, and this seems like a logical thing to display.

Elesant:
@allen-servedio I think that's a pretty cool idea. It will definitely improve the "content" side of the design for developers. Imagine you can view a well formatted and syntax highlighted Dockerfile right from the app. It would be nice if we can figure out a way together to keep things super simple while are introducing new contents as well :)

jeffdm:
Awesome. Taking a look!

styles/new-container.less
```diff
@@ -186,6 +186,12 @@
           overflow: hidden;
           text-overflow: ellipsis;
           white-space: nowrap;
+          text-decoration: underline;
+          &:hover {
```
jeffdm: Can we remove the hover-over styling for now?

src/ImageCard.react.js
```diff
@@ -67,7 +77,7 @@ var ImageCard = React.createClass({
         <div>
           <div className="namespace">{namespace}</div>
           <OverlayTrigger placement="bottom" overlay={<Tooltip>{this.props.image.name}</Tooltip>}>
```
jeffdm: Since it's redundant, should we change the Tooltip text to `View on DockerHub`?

src/ImageCard.react.js
```diff
@@ -40,6 +41,15 @@ var ImageCard = React.createClass({
     var $tagOverlay = $(this.getDOMNode()).find('.tag-overlay');
     $tagOverlay.fadeOut(300);
   },
+  handleRepoClick: function () {
+    var $repoUri = 'https://registry.hub.docker.com/'
```
jeffdm: + add semicolons

jeffdm:
Awesome! This is really cool (and asked for a lot). I added some comments

allen-servedio:
Sorry for the delay - I made the changes you suggested. Let me know if you want anything else done to it. Thanks for considering the change!

jeffdm:
LGTM, merging!

jeffdm:
Thanks for this! Awesome!

allen-servedio:
Hey Jeff - I goofed. I missed one of the tool tips (sorry about that). I pushed another commit to fix it (it is just below the one I fixed).

jeffdm:
No worries! I'll pull it in.

Thanks again @allen-servedio !

jeffdm:
Ok. Merged! Great! Thanks @allen-servedio !

@DieterReuter this is a neat idea, showing web pages in-app vs opening a separate browser window. We'd need some sort of navigation tree built-in, similar to iOS & Android apps. cc @Elesant

Issue #319, opened 2015-03-23
Running on windows


domderen:
Hi,

Could someone else please try to load the application on Windows? I am trying to start the app, and it does load. I managed to get the dependencies running, and I am not modifying the Setup scripts to download and install everything by itself. I prepared the dependencies powershell script which downloads the docker and docker-machine binaries, and modified the application code to use them. I also modified the code so it could find the VBoxManage binary and manage VirtualBox on windows. It is still missing a step of installing VirtualBox on it's own, so you need to install VirtualBox yourself, but other than that the application should start.

And now here comes the problem. When I run the application it seems that the app starts with two instances of react. I see two instances of 

```javascript
"Download the React DevTools for a better development experience: http://fb.me/react-devtools", source: c:\Users\dderen\Documents\GitHub\kitematic\node_modules\react\lib\React.js (143)
```

in the console log, and later when react tries to do anything connected with components it fails with error:

```javascript
"TypeError: Cannot read property 'firstChild' of undefined", source: file:///C:/Users/dderen/Documents/GitHub/kitematic/build/Main.js (47)
```

And the stacktrace points to react sources.

I debugged a bit more, and at application startup it seems that react is loaded twice, once by the kitematic app itself, and once by react-router, even though it was already loaded.

I tested this on Mac as well, and there is no issue, react is loaded only once, by the kitematic app. It is not loaded by react-router (I mean require command just takes the existing instance, instead of trying to initialize a new one). I also tested it on two different Windows boxes (one with Win 7 and one with Win 8.1), and the issue persists on both.

I would appreciate any help with this issue, and I have a problem even finding the root cause of it. I am not sure if it is connected with the Atom-Shell, or with react or with react-router or what...

Can anyone else test this setup on Windows as well to see if the behavior is reproducible globally?
My branch with windows changes is available at: https://github.com/domderen/kitematic/tree/win_client

Thanks in advance!

domderen:
Hey, the issue was solved by my friend @KamilSzot. He found out that the usage of 

```html
<script src="Main.js"></script>
```

in Index.html causes the issue. I changed the setup to be

```html
<script src="Startup.js"></script>
```

and added new file Startup.js which has the following content

```javascript
require('./Main');
```

And now the application works as expected. I will confirm that it also works on Mac, and prepare a pull request to fix the issue.

jeffdm:
Sorry for the delay in getting back. I've been following your work on Windows - this is awesome!

[...]

PR #329, opened 2015-03-24
A fix required to load only one instance of React in the application. 

domderen:
Without such fix, on windows application loads two instances of React at start time, and crashes.

This is connected to the issue #319

### DIFF
--- a/index.html
+++ b/index.html
@@ -6,6 +6,6 @@
     <title>Kitematic</title>
   </head>
   <body>
-    <script src="Main.js"></script>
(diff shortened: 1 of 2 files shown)


domderen:
Ultimately it is an issue with Atom-Shell raised here: https://github.com/atom/atom-shell/issues/1289 but I guess it might take them a while to fix...

Elesant:
Thank you so much @domderen for helping us with Windows :+1:

jeffdm:
LGTM. Merging

PR #359, opened 2015-04-03
Replace hard-coded error message with props.

### LINKED ISSUE #355: Kitematic shows fatal error when attempting to create a container from `scratch`
I go to create a new image from https://registry.hub.docker.com/_/scratch/

No VPN software running

image

This happens with a lot of other images as well, although some work just fine

koddsson:
Make sure that errors that happen when pulling image layers bubble up to the UI. We do this by catching errors coming from the docker client, calling the callback with that error and setting the global variable `_error` as the error we just caught. This is to make sure that when the stream ends we call the callback with the error as well as to not to over-ride it.

Finally we remove the hard-coded message in the UI and display the error that gets passed in to it through the props attribute. This is the only part I'm not super sure on if we really want since it must have been set there for a reason. If I can get a re-producible test case I can make sure that it's still passing with this patch.

Fixes #355.

Signed-off-by: Kristján Oddsson <koddsson@gmail.com>

### DIFF
--- a/src/ContainerHome.react.js
+++ b/src/ContainerHome.react.js
@@ -74,7 +74,9 @@ var ContainerHome = React.createClass({
     if (this.props.error) {
       body = (
         <div className="details-progress">
-          <h3>There was a problem connecting to the Docker Engine in the VirtualBox VM.<br/>This could be caused because this Mac is currently connected to a VPN, blocking access to the VM. If the issue persists, please <a onClick={this.handleErrorClick}>file a ticket on our GitHub repo.</a></h3>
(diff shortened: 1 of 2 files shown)


jeffdm:
Awesome! Thanks @koddsson :). This is a much asked for feature. Higher test coverage is a work in progress, so look out for that soon.

jeffdm:
Taking a look

jeffdm:
Just ran it locally. This looks fantastic. The code looks great too. LGTM

Merging.

koddsson:
Whooo! :balloon:

PR #363, opened 2015-04-05
Update tags on mounting ImageCard component.

### LINKED ISSUE #350: Cannot start any debian container.
I can start any other containers but Debian containers don't seem to be responsive. When created they just stop immediately and I cannot run or restart them. The log also stays empty.

koddsson:
Should fix #350 in the sense that the error "No command specified" will bubble up to the UI when trying to start a container that has no `CMD` in the Dockerfile.

I made example usage gifs using the `google/debian` image:

- [before](https://cloud.githubusercontent.com/assets/318208/6997501/d318902e-dbac-11e4-881a-f4610fd5e89a.gif)
- [after](https://cloud.githubusercontent.com/assets/318208/6997500/d2d69782-dbac-11e4-8cb7-a6ab25efc262.gif)

This does not fix the "No logs for this container" when booting containers such as the official ubuntu and debian images since they have `CMD` statements in their Dockerfile.

### DIFF
--- a/src/ImageCard.react.js
+++ b/src/ImageCard.react.js
@@ -31,11 +31,11 @@ var ImageCard = React.createClass({
   handleTagOverlayClick: function (name) {
     var $tagOverlay = $(this.getDOMNode()).find('.tag-overlay');
     $tagOverlay.fadeIn(300);
-    $.get('https://registry.hub.docker.com/v1/repositories/' + name + '/tags', function (result) {
+    $.get('https://registry.hub.docker.com/v1/repositories/' + name + '/tags', result => {
       this.setState({
         tags: result
       });
-    }.bind(this));
+    });
   },
   handleCloseTagOverlay: function () {
     var $tagOverlay = $(this.getDOMNode()).find('.tag-overlay');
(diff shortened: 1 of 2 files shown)


src/ImageCard.react.js
```diff
@@ -50,6 +50,14 @@ var ImageCard = React.createClass({
     }
     util.exec(['open', $repoUri + this.props.image.name]);
   },
+  componentDidMount: function() {
+    $.get('https://registry.hub.docker.com/v1/repositories/' + this.props.image.name + '/tags', function (result) {
+      this.setState({
+        tags: result,
+        chosenTag: result[0].name
+      });
+    }.bind(this));
```
jeffdm: Since we have es6 support, you can use fat arrow to avoid needing .bind(this)

- Remove .bind(this)
- Change:

```
function (result) {
}
```

to

```
(result) => {
}
```

jeffdm:
Awesome!! LGTM after that one comment

src/ImageCard.react.js
koddsson: I keep forgetting that there's ES6 support :eyeglasses:

jeffdm:
Great! Thanks for updating the old code as well. Merging

PR #364, opened 2015-04-05
Fall-back to image name as key when there is no id

koddsson:
When a container is first started and the image is downloading there is no id set. With this patch we'll check if the container has a id and and is downloading, if so we'll substitute it with the image name and render such.

This will fix the react key props warning in the developer console as well as maybe give us some sort of performance boost(?)

### DIFF
--- a/src/ContainerList.react.js
+++ b/src/ContainerList.react.js
@@ -9,8 +9,14 @@ var ContainerList = React.createClass({
   render: function () {
     var self = this;
     var containers = this.props.containers.map(function (container) {
+      var container_id = container.Id;
+      if (!container_id && container.State.Downloading) {
+        // Fall back to the container image name when there is no id. (when the
+        // image is downloading).
+        container_id = container.Image;
+      }
       return (
-        <ContainerListItem key={container.Id} container={container} start={self._start}/>
+        <ContainerListItem key={container_id} container={container} start={self._start} />
       );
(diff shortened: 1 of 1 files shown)


src/ContainerList.react.js
```diff
@@ -9,8 +9,14 @@ var ContainerList = React.createClass({
   render: function () {
     var self = this;
     var containers = this.props.containers.map(function (container) {
+      var container_id = container.Id;
```
jeffdm: containerId (camel case :)

jeffdm:
Thanks for the PR! Added a small comment.

src/ContainerList.react.js
koddsson: Sounds good! :)

koddsson: Also would you consider something like [jscs](http://jscs.info/) to automatically check those rules? (I've already added `requireCamelCaseOrUpperCaseIdentifiers` to my local `.jscsrc` ;)

jeffdm:
Any way to sign that last commit? Sorry about that.

Working on an easy way to sign once for all commits or similar.

koddsson:
I've forced pushed and added a pre-commit hook to remind me! :)

jeffdm:
Great! Sorry for the delay.

LGTM! Merging

PR #370, opened 2015-04-08
Updating Windows Branch with latest changes

jeffdm:


jeffdm:
I'll just merge this if it's okay @domderen since it's just upstream changes (and a little fix to change UID to be a number instead of a string in SetupUtil)

domderen:
Yeah all looks good. Does that add/remove any points from the windows client list?

jeffdm:
Just the point where these changes work on Mac as well (pretty much worked out of the box).

jeffdm:
Thanks for looking @domderen :)

PR #373, opened 2015-04-13
Fixing bug where logs would include headers

jeffdm:


Elesant:
LGTM! Thanks @jeffdm!

PR #383, opened 2015-04-16
Adding back custom URL for Hub-Button

jeffdm:
For further review


src/Main.js
```diff
@@ -27,14 +27,48 @@ setInterval(function () {
...
 router.run(Handler => React.render(<Handler/>, document.body));
 
+ipc.on('application:quitting', opts => {
+  if (!opts.updating && localStorage.getItem('settings.closeVMOnQuit') === 'true') {
+    machine.stop();
+  }
+});
+
+ipc.on('application:open-url', opts => {
+  var repoRegexp = /[a-z0-9]+(?:[._-][a-z0-9]+)*/;
```
jeffdm: Double checking this regexp

src/app.js
```diff
@@ -49,3 +59,27 @@ ipc.on('application:quitting', () => {
...
+
+ipc.on('application:open-url', opts => {
+  var parser = document.createElement('a');
+  parser.href = opts.url;
+
+  if (parser.protocol !== 'docker:') {
+    return;
+  }
+
+  var pathname = parser.pathname.replace('//', '');
```
diogomonica: It would be great to add comments with an example of a valid docker:// URL and what happens as we parse it.

src/utils/URLUtil.js
```diff
@@ -0,0 +1,44 @@
+var util = require('./Util');
+var parseUri = require('parseUri');
+var containerStore = require('../stores/ContainerStore');
+
+module.exports = {
```
jeffdm: @diogomonica here's the updated method with annotations, I pulled it out of the main app.js so I could unit test it. How does it look?

src/app.js
```diff
@@ -49,3 +55,15 @@ ipc.on('application:quitting', () => {
     machine.stop();
   }
 });
+
+// Event fires when the app receives a docker:// URL such as
+// docker://repository/run/redis
+ipc.on('application:open-url', opts => {
```
jeffdm: Entry point for URL generation.

src/utils/URLUtil.js
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    var parser = parseUri(url);
+
+    if (parser.protocol !== 'docker') {
+      return false;
+    }
+
+    // Get the type of object we're operating on, e.g. 'repository'
+    var type = parser.host;
```
diogomonica: Should add: if type not in type_whitelist, return

src/utils/URLUtil.js
```diff
@@ -0,0 +1,44 @@
...
+
+    // Separate the path into [run', 'redis']
+    var tokens = parser.path.replace('/', '').split('/');
+
+    // Get the method trying to be executed, e.g. 'run'
+    var method = tokens[0];
```
diogomonica: Should add: if method not in method_whitelist, return

src/utils/URLUtil.js
```diff
@@ -0,0 +1,44 @@
...
+
+    // Get the repository namespace and repo name, e.g. 'redis' or 'myusername/myrepo'
+    var repo = tokens.slice(1).join('/');
```
diogomonica: We don't seem to be checking for total number of tokens here. This could actually return a repo in the form of first/second/../fifth.

I think we need a regular expression here to enforce ^{first_group}/{second_group}$

src/utils/Util.js
```diff
@@ -52,5 +52,61 @@ module.exports = {
     } catch (err) {}
     return settingsjson;
   },
+  isOfficialRepo: function (name) {
+    if (!name || !name.length) {
+      return false;
+    }
+    var repoRegexp = /^[a-z0-9]+(?:[._-][a-z0-9]+)*$/;
```
diogomonica: Why does this regular expression distinguish an official repo from a non-official one? At least a comment is needed.

src/utils/URLUtil.js
```diff
@@ -0,0 +1,44 @@
...
+    // Only accept official repos for now
+    if (!util.isOfficialRepo(repo)) {
+      return false;
+    }
+
+    if (type === 'repository' && method === 'run') {
+      containerStore.setPending(repo, 'latest');
+      return true;
+    }
+  }
```
diogomonica: Shouldn't this explicitly return false at the end of the method?

diogomonica:
LGTM.

These changes make this code a lot more defensive. As long as remote disable works, I think we're in a good spot to stop anything that goes wrong.

Good work @jeffdm.

jeffdm:
@diogomonica Great, thanks for the review. Doing a sanity check then merging.

PR #385, opened 2015-04-16
Bumping to 0.5.14 with docker 1.6 and docker-machine 0.2.0

jeffdm:
- Also removing some bad practices in package.json where the `postinstall` would do some unnecessary dev setup.


jeffdm:
cc @Elesant

Elesant:
Awesome! LGTM!

PR #399, opened 2015-04-20
Fixes progress bar for old v1 images

jeffdm:
- New v2 images still have some issues
- Further work required to choose registry v2 by default (right now we chose v1) and then backfall to v1


Elesant:
LGTM!

jeffdm:
Thanks, merging!

PR #456, opened 2015-04-29
Added support for env shell called via settings

TeckniX:
Implemented a simple check for a custom shell assigned by the user within the Global settings.


jeffdm:
LGTM, awesome! Thanks @TeckniX!

PR #595, opened 2015-06-04
Search Infinite scrolling

TeckniX:
It seems that I've encountered some duplicates from time to time when parsing results from the API.

More than an infinite scroll, this is a lazy loading method, where pages after pages are requested and added to the search results. 

Hope you like it :)


src/components/InfiniteGrid.js
```diff
@@ -0,0 +1,224 @@
+var React = require('react/addons');
+var PureRenderMixin = React.addons.PureRenderMixin;
+
+/**
+ * Inspired from: https://github.com/ggordan/react-infinite-grid
```
jeffdm: How close is this code to the inspiration? Should we add the MIT license in?

src/components/InfiniteGrid.js
```diff
@@ -0,0 +1,224 @@
...
+        width: this._itemWidth(),
+        gridWidth: 0,
+        itemsPerRow: 2,
+      },
+    };
+  },
+
+  // METHODS
+
+  _wrapperStyle: function() {
```
jeffdm: Could we move this to less?

Eventually I'm hoping to have all styles in Javascript, like this, but I think consistency is important.

src/components/InfiniteGrid.js
```diff
@@ -0,0 +1,224 @@
...
+    lazyLimit = lazyLimit < 1? 1:lazyLimit;
+    if (!this.state.initiatedLazyload && (this.state.maxItemIndex >= lazyLimit) && (this.state.maxItemIndex < this.props.maxEntries) && this.props.lazyCallback) {
+      this.setState({initiatedLazyload: true });
+      this.props.lazyCallback(this.state.maxItemIndex);
+    } else if (this.state.maxItemIndex === this.props.maxEntries){
+      this.setState({initiatedLazyload: false });
+    }
+  },
+
+  // LIFECYCLE
```
jeffdm: Not sure if we need this comment.

src/utils/RegHubUtil.js
```diff
@@ -47,8 +48,11 @@ module.exports = {
       let repos = _.map(data.results, result => {
         return this.normalize(result);
       });
+      let pageLimit = data.num_pages;
+      // Error in API, more items due to array count
+      let maxResults = data.num_results-1;
```
jeffdm: nitpick, spacing: `data.num_results - 1;`

src/utils/RegHubUtil.js
```diff
@@ -37,7 +38,7 @@ module.exports = {
 
     searchReq = request.get({
       url: 'https://registry.hub.docker.com/v1/search?',
-      qs: {q: query, page}
+      qs: {q: query, page: page, time: +new Date()}
```
jeffdm: What does the `+` do here?

src/utils/RegHubUtil.js
```diff
@@ -25,7 +26,7 @@ module.exports = {
     return obj;
   },
 
-  search: function (query, page) {
+  search: function (query, page=1) {
```
jeffdm: Great use of default arguments. I had no idea this existed in ES6. COOL!

src/components/InfiniteGrid.js
TeckniX: A lot of scroll/lazycallback inspiration came from it, but I did remove a ton of logic, updated broken items and overall association to work for Kitematic. I don't mind adding the MIT license if needed

jeffdm:
This looks AMAZING. Thank you @TeckniX . Great example of a contribution that touches every part of flux.

Taking a look!

jeffdm:
RE the js styles => less comment. I just realized the styles are dynamic - so having them in JS makes sense.

src/utils/RegHubUtil.js
TeckniX: `+new Date()` creates a date object that is then coerced into a number (same as `Date.now()`)  - I only have it to make sure cached response isn't seen, but could be removed, or `Date.now()` can be used in its place.

src/components/InfiniteGrid.js
```diff
@@ -0,0 +1,224 @@
...
+    let entries = this.props.entries;
+    let loading = (
+      <div className="spinner la-ball-clip-rotate la-dark la-lg"><div></div></div>
+      );
+
+    return(
+      <div ref="wrapper" className="infinite-grid-wrapper" onScroll={this._scrollListener} style={this._wrapperStyle()}>
+        <div ref="grid" className="infinite-grid result-grid" style={this._gridStyle()}>
+          {entries}
+          { this.state.initiatedLazyload ? loading : null }
```
jeffdm: Nitpick: no spaces between `{`  `}` like above

src/components/NewContainerSearch.react.js
```diff
@@ -2,6 +2,8 @@ var _ = require('underscore');
 var React = require('react/addons');
 var Router = require('react-router');
 var RetinaImage = require('react-retina-image');
+//var Infinite = require('react-infinite');
```
jeffdm: Remove comment

src/utils/RegHubUtil.js
TeckniX: Haha, tbh I threw it in there out of coding habit and was pleased that it worked - Totally forgot regular JS doesn't like it.

src/components/InfiniteGrid.js
jeffdm: Wow. Impressive :smiley:

I think this inspiration link is ok for now, maybe just add `(MIT LICENSE)` to the comment

src/components/InfiniteGrid.js
```diff
@@ -0,0 +1,224 @@
...
+    if (!this.props.wrapperHeight) {
+      this.setState({
+        wrapperHeight: window.innerHeight,
+      });
+    }
+    this._updateItemDimensions();
+    this._visibleIndexes();
+  },
+
+  // RENDER
```
jeffdm: Comment necessary?

src/components/InfiniteGrid.js
```diff
@@ -0,0 +1,224 @@
...
+    if (typeof this.props.renderRangeCallback === 'function') {
+      this.props.renderRangeCallback(this.state.minItemIndex, this.state.maxItemIndex);
+    }
+  },
+
+  componentWillUnmount: function() {
+    window.removeEventListener('resize', this._resizeListener);
+  },
+
+  // LISTENERS
```
jeffdm: Comment necessary?

jeffdm:
Some comments added, and LGTM after that. Again - great add @TeckniX .

@mchiang0610  @Elesant can you LGTM?

TeckniX:
As a side note there's a good amount of code that could be removed from the InfiniteGrid.js, as it's not needed/used. 

One thing that may needed some tweaking is the 'Other Result' element height - I don't have any repo within my DockerHub login, so it may look different based on that.

src/utils/RegHubUtil.js
```diff
@@ -47,8 +48,11 @@ module.exports = {
       let repos = _.map(data.results, result => {
         return this.normalize(result);
       });
+      let pageLimit = data.num_pages;
+      // Error in API, more items due to array count
+      let maxResults = data.num_results - 1;
       if (response.statusCode === 200) {
-        repositoryServerActions.resultsUpdated({repos});
+        repositoryServerActions.resultsUpdated({repos, pageLimit, maxResults});
       }
```
jeffdm: Often `error` is `null`, but `responseCode` may be 500.

Maybe add an `else` statement that calls `repositoryServerActions.error({error:data.detail});`?

src/components/InfiniteGrid.js
```diff
@@ -0,0 +1,216 @@
+var React = require('react/addons');
```
jeffdm: Rename this file to InfiniteGrid.react.js

jeffdm:
Ok, :smiley: best to remove any unused code so we don't need to maintain it later

mchiang0610:
Just ran it. Super cool. 

A couple comments in the screenshot: 

image

TeckniX:
I removed the inner-wrapper so that the entire page scrolls as it did before - I did however have to pass the 'parent' component down to the InfiniteGrid child, so that I could figure out the height of it, and calculate when the lazyLoad should happen. 

My lack of React experience shows here, as I grossly passed the entire `this` object - Feel free to suggest a better way of doing this :)

jeffdm:
@TeckniX should be okay for now. Code LGTM! Thanks for the fast fix :smiley_cat: 

@mchiang0610 can you try again?

mchiang0610:
@TeckniX the loading spinner should be at the bottom. 

@jeffdm there are performance issues in Kitematic when many repos are listed. (Kitematic issue)

jeffdm:
Rendering the little results card causes lag. I think it's from the gradients. I'll check tomorrow

TeckniX:
I've actually run into something similar while trying to implement 'tag selection/prompt' during creation. 
I believe the issue is that every card is creating a listener for the `tagStore` change, but 1 `tagStore` change will trigger all cards to capture the update. 
As the number of cards grows, so does the listener, yet they all take the same action. (you can throw a debug in the `update` function for the `ImageCard` and see it being triggered over and over. 

If you'd like to discuss an `ImageCard` component redesign, then I'm all about it - The listener needs to be in the parent component I believe to prevent the multi-listeners.

jeffdm:
Sorry for the delay on this @TeckniX cc @FrenchBen

Good catch on that on ImageCard. I feel we should refactor that before merging this, and I think things will be pretty smooth.

FrenchBen:
There's def a few improvements we can make - Do you have any specific ideas as to how you want to handle it?

jeffdm:
@FrenchBen A small change would be to `listen` only when the Tag list shows, and unlisten when it's closed. Just a quick thought, will need to experiment with the performance tools

FrenchBen:
#closing for lack of updates and the need to refactor a lot of the listeners for this to work properly

PR #597, opened 2015-06-05
Fixing setting up custom shell in container terminal

VojtechBartos:
Hey guys,
i wanted to set custom shell to `bash` instead of `sh` which is default. I set environment `ENV SHELL /bin/bash`, but it was still picking up default option `sh` every time. I have found in code when you trying to check if somebody wants to set custom shell that you are accessing to array like to object, so i have added simple reduce function to make it actually object.

Hopefully everything is ok ;)

Let me know if you have any questions.
Vojta

Signed-off-by: Vojta Bartos <hi@vojtech.me>

### DIFF
--- a/src/components/ContainerDetailsSubheader.react.js
+++ b/src/components/ContainerDetailsSubheader.react.js
@@ -102,7 +102,11 @@ var ContainerDetailsSubheader = React.createClass({
     if (!this.disableTerminal()) {
       metrics.track('Terminaled Into Container');
       var container = this.props.container;
-      var shell = ContainerUtil.env(container).SHELL;
+      var shell = ContainerUtil.env(container).reduce((envs, env) => {
+        envs[env[0]] = env[1];
+        return envs;
+      }, {}).SHELL;
+
       if(typeof shell === 'undefined') {
         shell = 'sh';
       }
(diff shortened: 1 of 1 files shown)


src/components/ContainerDetailsSubheader.react.js
```diff
@@ -102,7 +102,11 @@ var ContainerDetailsSubheader = React.createClass({
     if (!this.disableTerminal()) {
       metrics.track('Terminaled Into Container');
       var container = this.props.container;
-      var shell = ContainerUtil.env(container).SHELL;
+      var shell = ContainerUtil.env(container).reduce((envs, env) => {
+        envs[env[0]] = env[1];
+        return envs;
+      }, {}).SHELL;
+
       if(typeof shell === 'undefined') {
```
jeffdm: Would it be better to just do `if (!shell)` ?

VojtechBartos: @jeffdm yeah, definitely, but i wanted to do just necessary steps to make it work. Anyway, i have changed it.

jeffdm:
Thanks for the PR! Code LGTM

@mchiang0610 can you test functionality

mchiang0610:
@jeffdm anything specific to run it with?

jeffdm:
Not much to test-  containers with $SHELL = `bash` should start with bash now instead of `sh`.

Merging

PR #616, opened 2015-06-11
Add small notification

TeckniX:
*Functionality*
Small addition to the current Kitematic Header, will allow for notifications to be shown (probably need some UI inputs)

*UI UPDATE*
Applied some CSS changes to prevent the header from being squished when size of logs is bigger than its container.

notification


TeckniX:
Will need to fix this too - On-Hold till fix

mchiang0610:
@TeckniX When I click create on an image, the view no longer changes to the container. Is this part of this PR or a side effect? 

What are the states for the notifications? (What errors / logs will trigger the notifications?)

TeckniX:
I had botched the previous PR and had to clean it up - The new PR should only have 2 file changed.
The notification is triggered when the ContainerStore Error is set. This is done via a simple listener during the ComponentDidMount

TeckniX:
The UI look and feel:
kitematic-notification

From the new UI looks like anything in the header-area couldn't be clicked, as there are some index issues; this forced the notifications to be pushed down next to the tabs

FrenchBen:
#closing due to recent release updates which clarify some of the errors and allow copy&paste

Issue #535, opened 2015-05-26
Container logs inside freeze after stopping and restarting the container 


mchiang0610:

TeckniX:
Is this on a particular container? The logs are being displayed properly on my `ghost` container:

```logs
> ghost@0.6.2 start /usr/src/ghost
> node index

Migrations: Up to date at version 003
Ghost is running in development... 
Listening on 0.0.0.0:2368 
Url configured as: http://localhost:2368 
Ctrl+C to shut down

> ghost@0.6.2 start /usr/src/ghost
> node index

Migrations: Up to date at version 003
Ghost is running in development... 
Listening on 0.0.0.0:2368 
Url configured as: http://localhost:2368 
Ctrl+C to shut down
```

ianbytchek:
Not sure if I get the exact same thing. After restarting containers logs are not updating properly, I have to switch between home and logs tabs to get the refresh.

mchiang0610:
@TeckniX Sorry for the delay. This seems to be a sporadic problem. I cannot reproduce most of the time. Opened an issue so we are aware. 

@ianbytchek Yes, same error.

ianbytchek:
It happens all the time. Start the container that prints something into stdout, restart it, logs stop automatically refreshing, only by switching from home and logs tabs or vice versa.

TeckniX:
@ianbytchek thanks for the good step-by-step, I now see it too. I'll start working on a fix

FrenchBen:
Closing as this has been fixed in recent versions

PR #622, opened 2015-06-12
Fixed stale log refresh

TeckniX:
Previous logs only updated when the component mounted, this fix allows for logs to be refreshed whenever the actual container is updated, usually triggered by `containerServerActions.updated({container})`

This should fix issue https://github.com/kitematic/kitematic/issues/535

### DIFF
--- a/src/stores/ContainerStore.js
+++ b/src/stores/ContainerStore.js
@@ -3,6 +3,7 @@ import deepExtend from 'deep-extend';
 import alt from '../alt';
 import containerServerActions from '../actions/ContainerServerActions';
 import containerActions from '../actions/ContainerActions';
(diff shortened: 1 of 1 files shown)


mchiang0610:
Thanks @TeckniX. Trying it out

mchiang0610:
@jeffdm LGTM. Can you review the code? 

Thanks!

jeffdm:
LGTM

This is a great fix. I think later we should move LogStore to the flux architecture.

TeckniX:
@jeffdm I concur with moving LogStore to the flux architecture, although it was a bit more involved than this quick fix and thus put on the back-burner :P

PR #654, opened 2015-06-17
Added support for docker tty stdin options

TeckniX:
I've been using busybox-like data containers and noticed that for some it was needed to add `docker run -ti` options to have a pseudo-tty allocated to the container otherwise the container wouldn't stay running. 

Since `-i -t` are commonly added together, it was only normal to also support that option. 

Look and feel:
kitematic screen shot

### DIFF
--- a/src/components/ContainerSettingsGeneral.react.js
+++ b/src/components/ContainerSettingsGeneral.react.js
@@ -22,7 +22,7 @@ var ContainerSettingsGeneral = React.createClass({
       return [util.randomId(), e[0], e[1]];
     });
 
-    let [tty, openStdin] = ContainerUtil.mode(this.props.container) || [false, false];
+    let [tty, openStdin] = ContainerUtil.mode(this.props.container) || [true, true];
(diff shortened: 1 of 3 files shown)


Elesant:
This is so awesome! <3

mchiang0610:
@TeckniX This is HUGE. Tried running it, and it works very well. 

I think the attach standard streams to a tty, including stdin if it's not closed should be on by *default*. 

Can we change the "Foreground Options" text to "Advanced Options"? The reason is so users learning Docker / Kitematic will not try to break things if they don't know.

jeffdm:
Agreed^ This is awesome! Thanks @TeckniX !

I agree with Mike's comment on having this on default since it's so powerful. Best to do this in the `DockerUtil.js` > `createContainer` method.

TeckniX:
I'll go ahead and update the verbiage and can set it as default during the createContainer

jeffdm:
LGTM. Merging

Issue #376, opened 2015-04-15
Allow mounting custom volume directories


jeffdm:
Right now only volumes defined in `Dockerfiles` can be host-mounted via Kitematic. Users should have a way to host-mount into arbitrary folders in the container.

kematzy:
+100

Also, some kind of documentation on how to convert a great docker container into a fully supported Kitematic container with custom volume addition, logs and previews would be great too.

jeffdm:
@kematzy great! Thanks for the note!

gregorydickson:
+1

bauschan:
+1

Blux:
+1

jb510:
+1000 on @kematzy's  "documentation on how to convert a great docker container into a fully supported Kitematic container with custom volume addition, logs and previews would be great too"

I've been banging my head into a lot of glass doors today trying to figure that out on my own.  

The only container that seems to work with mounted volumes is kitematic/hello-world-nginx and I can't even find the source and how how it does it.  

Every other container with volumes breaks as soon as I enable them.

jeffdm:
@jb510 really sorry to hear that! It's mostly from the Dockerfile. The summary:

Volumes: Make sure there's a `VOLUME` directive in the Dockerfile
Ports: Make sure there's an `EXPOSE` directive in the Dockerfile
Environment variables:  Use `ENV` directives in the Dockerfile

 and we'll post a doc article on the website (kitematic.com/docs) soon

jb510:
@jeffdm Oh goodness don't be sorry! Kitematic is freaking amazing...  All sorts of possibilities are swirling through my head.  

One question.  Where do I find the Dockerfile for kitematic/hello-world-nginx ? I'd like to read/understand it and I can't find it.

jeffdm:
@jb510 here https://github.com/kitematic/html although we had to squash the image for faster download.

@mchiang0610 it may be a good idea to make this an automated build instead for the Dockerfile reference

jb510:
TY!  Exactly what I was looking for!

ajermakovics:
+1

nym:
+1

mchiang0610:
This is definitely something that should be on the roadmap for Kitematic. Is anyone willing to help us work on this as we prioritize on the Windows version of Kitematic in the meantime? @kematzy @ajermakovics @jb510 @nym 

Would love to hear thoughts on how it should be implemented.

alexj:
[...]

PR #721, opened 2015-06-25
Implement a configurable 'Debug' tab on containers

zedtux:
This pull request add a new option

screenshot from 2015-06-25 20 07 13

Which then add a new tab (when debug enabled) which shows the JSON representation of the selected container at the moment you open the tab:

screenshot from 2015-06-25 20 08 36

I needed this as I'm implementing the Volumes management in order to fix issue #376.

### DIFF
--- a/src/actions/ContainerActions.js
+++ b/src/actions/ContainerActions.js
@@ -39,6 +39,12 @@ class ContainerActions {
   run (name, repo, tag) {
     dockerUtil.run(name, repo, tag);
   }
+
+  inspect (name, component, callback) {
+    dockerUtil.inspectContainer(name, function (result) {
+      callback(result, component);
+    });
+  }
 }
 
 export default alt.createActions(ContainerActions);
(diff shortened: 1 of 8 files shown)


TeckniX:
@zedtux Although I really like this option/idea, I think most users would be confused by what is actually represented there. IMHO Kitematic is in place to allow rapid setup without having to read the API doc and understand what you've shown on that tab. Maybe enable debugging is actually in the Dev Tools?

More than a debug an actual list/table of the container settings would be more useful. 
Keep in mind that if the feature/setting isn't something you can change via Kitematic, then it's useless as the user will have gone through the command-line anyway to see it.

zedtux:
The users will never see this tab until he intentionally enable it in the options.
This is a very first version, developed in 10 minutes. It will need obviously improvements like showing useful information, and allow specific actions that only developers would like to use.

You can see a parallel with Safari where initially you don't have the developer tools enabled, so that you cannot inspect a page. Then going to the settings, and enabling the option add the new contextual menu entry to inspect a page.

In my opinion, it will not confuse users, as in the case they enable the option, they do it on purpose.

FrenchBen:
#closing as most users would use the CLI if they are doing some debugging

PR #796, opened 2015-07-08
Migrated all required module calls to ES6 imports equivalents

TeckniX:
A few duplicate imports were present and thus removed during this refactor. 


jeffdm:
LGTM, thanks!

Issue #755, opened 2015-07-03
Docker CLI with zsh doesn't work on mac os x 


vincent-herlemont:
Hello,

I try kitematic and docker for the first time but I can't push an sample repository in the hub.
Maybe the button "DOCKER CLI" on left dow has was created for this use.
But when I click on, this open terminal but the casual command appear.
I use zsh for my Terminal, it's maybe the problem.

How can get around this issues ?

jeffdm:
If you press the `Docker CLI` button and then type `docker ps`, does it work?

vincent-herlemont:
It's freak, see the small test in [this video](https://www.youtube.com/watch?v=gNZuxAHsTeg) ( two shell opens )

jeffdm:
@vincent-herlemont got it - definitely a bug. Thanks!

jeffdm:
@vincent-herlemont I'm having some trouble reproducing it the zsh bug but it does open two terminals.

Can you provide me the entire command (scroll up after hitting the `Docker CLI` button)?

vincent-herlemont:
Ho yes ! sorry, I'don't see that.

Here the command 
```
Last login: Sat Jul  4 09:10:32 on ttys000
bash -c "export PATH='/Users/vincentherlemont/Downloads/Kitematic (Beta).app/Contents/Resources/resources:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin' && clear && DOCKER_HOST=tcp://192.168.99.100:2376 DOCKER_CERT_PATH=/Users/vincentherlemont/.docker/machine/machines/dev DOCKER_TLS_VERIFY=1 /bin/zsh"
--- ~ » bash -c "export PATH='/Users/vincentherlemont/Downloads/Kitematic (Beta).app/Contents/Resources/resources:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin' && clear && DOCKER_HOST=tcp://192.168.99.100:2376 DOCKER_CERT_PATH=/Users/vincentherlemont/.docker/machine/machines/dev DOCKER_TLS_VERIFY=1 /bin/zsh"
```

vincent-herlemont:
Don't forget this bug please :pray: !

tiramisuconpanna:
I seem to have a similar problem: I have previously installed boot2docker and I'm also using oh-my-zsh. When I click on Docker-Cli, I just get to my default terminal. When I use docker, it is connecting to boot2docker, and not the kitematic vm

[...]

PR #827, opened 2015-07-17
Fixes error of docker CLI not working under zsh

phersh:
This commit fixes issue #755. zsh is a common shell alternative on OS X but the terminal script assumes the bash shell. I made a few simple changes to make the script work in both bash and zsh.

### DIFF
--- a/resources/terminal
+++ b/resources/terminal
@@ -2,7 +2,7 @@
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 PATH=$DIR:$PATH
-CMD="export PATH='$PATH' && clear && $*"
+CMD="export PATH='$PATH' &&clear"
 
 ITERM_EXISTS=`osascript <<EOF
 set doesExist to false
@@ -21,14 +21,14 @@ function open_iterm () {
       tell the first terminal
         launch session "Default Session"
         tell the last session
-          write text "bash -c \"$CMD\""
+          write text "$CMD"
         end tell
       end tell
     on error
       tell (make new terminal)
         launch session "Default Session"
         tell the last session
-          write text "bash -c \"$CMD\""
+          write text "$CMD"
         end tell
       end tell
     end try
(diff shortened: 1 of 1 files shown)


FrenchBen:
I actually use ZSH on my OSX and never had any issues - the 'bash' command you see there is just to run the docker export style commands, it should load your default shell otherwise.

Can you explain the issue this fixes?

phersh:
The terminal script doesn't correctly set the path. If you already have the path to the command line tools set from another installation of docker, or if you manually edited your bashrc or zshrc to add the path, then you won't see the error. You can easily check by removing the command line tools location from your path and then running kitematic and CL tools button.

FrenchBen:
What PATH are you referring to? CLI are installed in `/usr/local/bin`, which I thought all shell used?

xyzza:
I have the same bug under osx + zsh

FrenchBen:
I should perhaps ask - Is this a true ZSH or a terminal selected ZSH?
Aka - usermodded to use ZSH or you went to the Terminal -> Pref window and did "Run Command" under the 'shell' tab?

xyzza:
Yes, I wasn't totally correct, sorry. I should specify: It is Iterm2 + oh-my-zsh (I suppose it's second option with terminal selected zsh?)
Btw, after I installed boot2docker, the "docker" command appeared in CLI. I checked - it refers to kinematic VM (wich is dev) and it works fine now.
But without boot2docker there were no "docker" command at all, and that's kinda funny...

FrenchBen:
@xyzza By default Kitematic doesn't install the docker CLI - you have to 'tell it' to do so by going to File -> Install Docker Commands

xyzza:
@FrenchBen oh, that is the thing! It wasn't so obvious for me, thanks!

FrenchBen:
@xyzza  :+1:

FrenchBen:
Tried this patch with the latest master branch and it didn't work for me - @phersh Would you mind checking?
Thanks

phersh:
Hi @FrenchBen I'm working on it.

FrenchBen:
#closing in favor of recent releases and the fact that the current kitematic release works perfectly in my zsh

PR #834, opened 2015-07-17
Added support for DNS and DNS Search

FrenchBen:
Add support for custom DNS and Search domains within the advance tab.

### DIFF
--- a/src/components/ContainerSettingsAdvanced.react.js
+++ b/src/components/ContainerSettingsAdvanced.react.js
@@ -5,7 +5,7 @@ import ContainerUtil from '../utils/ContainerUtil';
 import containerActions from '../actions/ContainerActions';
 
 var ipRegex = /\b(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\b/;
-var domainRegex = /([a-z0-9-]+\.(?:com|net|org|co\.uk|localhost|dev))(?:\/|$)/;
+var domainRegex = /^([a-z0-9-]+(?:\.[a-z0-9-]+)+)(?:\/|$)/;
 
 var ContainerSettingsAdvanced = React.createClass({
   mixins: [React.addons.LinkedStateMixin],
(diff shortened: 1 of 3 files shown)


jeffdm:
Code LGTM

@mchiang0610 can you take a look?

mchiang0610:
If I run the demo nginx container in Kitematic and change the DNS to 8.8.8.8 (Google's DNS), the preview screen goes empty (correct behavior since it should not contact outside) but when I click open browser icon (or anywhere within), I get this error: 

>> [9972:0828/163650:INFO:CONSOLE(59)] "Uncaught TypeError: Error processing argument 0.", source: /Users/mc/github/kitematic/build/components/ContainerHomePreview.react.js (59)

@FrenchBen

FrenchBen:
@mchiang0610 I'll check to see what's going on with it - It shouldn't affect the preview screen as it's only modifying what's in the resolv.conf - I'll do some debugging. 

Thanks for the QA.

FrenchBen:
Cleaned up a lot of the Docker util start options - By default *no* option is allowed on 'start' but the dockerode module does allow the over-write on start of the container's settings. 

I left Bind in there, as I believe it was there originally, but it should really be moved to the createcontainer method.

src/components/ContainerSettingsAdvanced.react.js
```diff
@@ -1,8 +1,12 @@
+import _ from 'underscore';
 import React from 'react/addons';
 import metrics from '../utils/MetricsUtil';
 import ContainerUtil from '../utils/ContainerUtil';
 import containerActions from '../actions/ContainerActions';
 
+var ipRegex = /\b(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\b/;
+var domainRegex = /([a-z0-9-]+\.(?:com|net|org|co\.uk|localhost|dev))(?:\/|$)/;
```
fd: I think this is overly restrictive. The web is not limited to just `.com`,  `.net`, `.org`, `.co.uk`, `.localhost` and `.dev`.
Maybe something like `/^([a-z0-9-]+(?:\.[a-z0-9-]+)+)(?:\/|$)/` is be better.

FrenchBen: Good suggestion, although a lot looser, but should work nicely - I'll make the proper updates and will push it.

FrenchBen:
#closing due to lack of interest in DNS settings

PR #843, opened 2015-07-20
Added proper about page with tools versions

FrenchBen:
Will probably need some better banner graphic and design input from @Elesant
This will help provide an homogenous 'About' page for all kitematic platforms.

Preview:
kitematic-version

Signed-off-by: FrenchBen <me@frenchben.com>


src/menutemplate.js
```diff
@@ -17,7 +17,13 @@ var MenuTemplate = function () {
       submenu: [
       {
         label: 'About Kitematic',
-        selector: 'orderFrontStandardAboutPanel:'
```
timfallmk: I think we need a switch here. The `selector` is necessary for OSX.

FrenchBen: Not necessary, as it's commented out :P - I should delete it completely to prevent any further confusion.

timfallmk: Sorry, I meant it _is_ necessary for OS X. The switch would be to check with arch you're on. ;)

cclauss:
This looks great!  Really useful.

FrenchBen:
Latest changes include @Elesant's designs
about-cartoon

timfallmk:
:+1:

Elesant:
This is awesome! Let's merge it :)

cclauss:
Beautiful!!

PR #845, opened 2015-07-22
Fix volume data path on guide

tellel:
volume data path should be `~/Documents/Kitematic/<container's name>/`.


jeffdm:
Thank you!

PR #729, opened 2015-06-27
Add support for new version of iTerm2

xream:
FYI: [https://iterm2.com/applescript.html](https://iterm2.com/applescript.html)

PR #855, opened 2015-07-23
Add support for launching iTerm 2.9+

kudos:
The AppleScript bridge changed at some point since 2.1. While marked as
beta, the 2.9 series is the author's recommended version to use.

The two bridges are incompatible, requiring another function and version
checks.

### DIFF
--- a/resources/terminal
+++ b/resources/terminal
@@ -16,22 +16,49 @@ EOF`
 function open_iterm () {
   osascript > /dev/null <<EOF
   tell application "iTerm"
-    activate
-    try
-      tell the first terminal
-        launch session "Default Session"
-        tell the last session
-          write text "bash -c \"$CMD\""
-        end tell
-      end tell
-    on error
-      tell (make new terminal)
-        launch session "Default Session"
-        tell the last session
-          write text "bash -c \"$CMD\""
-        end tell
-      end tell
-    end try
+    if version < 2.9 then
+      activate
+      try
+      	tell current window
+          create tab with default profile
+    	    tell the current session of current window
+            write text "bash -c \"$CMD\""
+          end tell
+      	end tell
+      on error
+    	  try
+          tell the first terminal
+            launch session "Default Session"
+            tell the last session
+              write text "bash -c \"$CMD\""
+            end tell
+          end tell
+        on error
+          tell (make new terminal)
+            launch session "Default Session"
+            tell the last session
+              write text "bash -c \"$CMD\""
+            end tell
+          end tell
+        end try
+      end try
+    end if
+  end tell
+EOF
+}
+
+function open_iterm2point9 () {
+  osascript > /dev/null <<EOF
+  tell application "iTerm"
+  	if version ≥ 2.9 then
+      activate
+  		tell current window
+  			create tab with default profile
+  			tell first session of current tab
+  				write text "bash -c \"$CMD\""
+  			end tell
+  		end tell
+  	end if
   end tell
 EOF
 }
@@ -48,7 +75,7 @@ EOF
 }
 
 if [ "$ITERM_EXISTS" == "true" ]; then
-  open_iterm "$@" || open_terminal "$@"
+  open_iterm2point9 "$@" || open_iterm "$@" || open_terminal "$@"
 else
   open_terminal "$@"
 fi


kudos:
I think the `ITERM_EXISTS` stuff is even less required with this patch, happy to remove that too.

FrenchBen:
@kudos is this on the nightly build? We previously saw some issues with the nightly, see:
https://github.com/kitematic/kitematic/pull/729

resources/terminal
```diff
@@ -16,26 +16,53 @@ EOF`
...
+      end try
+    end if
   end tell
 EOF
 }
 
+function open_iterm2point9 () {
+  osascript > /dev/null <<EOF
+  tell application "iTerm"
+	if version ≥ 2.9 then
```
FrenchBen: Can you fix some of the indentation here?

kudos:
@FrenchBen it has recently entered [beta](https://iterm2.com/version3.html), which is what I'm running.

kudos:
@FrenchBen Just to be clear, the issue talked about in that other thread is not what's going on here, I'm running a beta not a nightly. iTerm2's Applescript has been [deliberately changed in a backwards-incompatible way](https://iterm2.com/applescript.html).

kudos:
I've also noticed that if my prompt doesn't come up quick enough the Applescript writes the command too soon and ends up being ignored. This is in the default Terminal.app along with iTerm2 stable and beta. Adding a delay of 1 second between creating a new tab/window and issuing the command solves it for me. Ironically, `eval "$(docker-machine env default)"` makes up ~25% of my shell startup time.

pzgz:
Looking forward to get this PR merged. THanks.

FrenchBen:
@pzgz We just need the stable release of 2.9 to be out and this will get added.

pzgz:
Looks like the PR itself has backward compatibility for iterms2 < 2.9. Should be ok even 2.9 is still in beta.

FrenchBen:
@pzgz normally I would agree but there could still be some changes to the software where this patch wouldn't apply. 
The goal is to have only 1 PR to fix it and not a bunch of follow up PR to keep fixing issues due to Nightly/Beta changes.

pzgz:
@FrenchBen Understood, thanks.

FrenchBen:
@kudos To keep the list of PR clean I'm going to close this until iTern has the stable release out - Thank you for providing this awesome patch; chances are we'll re-use it in the future when iTerm 2.9 is released

kudos:
@FrenchBen thanks for keeping me in the loop. I'm manually applying this patch each time I update Kitematic in the mean time.

FrenchBen:
@kudos no problem - I'll make sure to bring this back to life. I use iTerm on the daily (just not the nightly) so it will be pretty obvious when it breaks. :P

grempe:
If this PR is stable and works against both the older and newer iTerm2 I would vote for applying it now.  It may be entirely possible that iTerm2 will sit in beta for many months before a new 'stable' version is released. This beta was a surprise release after 18 months of development.

http://iterm2.com/version3.html

In the meantime we all get stuck with ugly little terminal windows.  Please handle both old and new versions of iTerm2.  You can remove the dual handler code later.

grempe:
Please also note that the developer of iTerm2 has provided sample code for supporting simultaneously supporting the AppleScript of both the old and the new versions here:

http://iterm2.com/applescript.html

petems:
I agree with @grempe, there seems to be a way to support both, and it takes a long time for iterm releases to become stable. Can we get this re-looked at?

FrenchBen:
As previously mentioned the idea behind the delay was to prevent adding code on nightly that was possibly never going to make it to beta/RC - Seeing the behavior in Beta is a good indication that it will make it to RC and release - I've re-opened this PR, but would like other people to try it if possible?

@grempe and @petems could you replace your `terminal` file with the one in this PR?
Should be as easy as Opening `/Applications/Docker Toolbox/Kitematic (Beta).app/Contents/Resources/resources` and replacing the `terminal` file with the one from kudos.

Please provide feedback here.

grempe:
I copied the terminal file as found here:

https://github.com/kudos/kitematic/blob/iterm2point9/resources/terminal

and it worked like a charm for me.  I am using iTerm2 version `Build 2.9.20160206`

Kudos to @kudos

FrenchBen:
:+1: Just need @petems to also confirm and I'll get it merged.

kudos:
@FrenchBen it's been in beta since I opened the PR, btw.

petems:
@FrenchBen Yep, just replaced the file with the one from this PR, works for me! :smile:

FrenchBen:
Thanks everyone for your patience, QA and above all thanks to @kudos for the work. 
Merging..

petems:
Awesome :dancer:

gnachman:
What should I suggest users to do? I'm preparing to release the stable version of iTerm2 3.0 soon. I want to help them upgrade incompatible programs. iTerm2 will detect when version 1.3.0 of the docker quickstart terminal is installed and give them upgrade instructions. Not sure what to say, though—I tried installing the docker toolbox today and it still had the old version of the script.

FrenchBen:
Thanks for chiming in @gnachman and also for being considerate enough to have a check in place for this. 
I checked the latest release of kitematic and it has the above merged code. The quick-start terminal may need to be tweaked slightly, which is on the toolbox side of things. It looks like there's an issue opened for this, I'll try to breath some life into it internally:
https://github.com/docker/toolbox/issues/348

gnachman:
@FrenchBen Any ETA on when this release of Kitematic will be in the Docker Toolbox? 1.11.1 seems to be the old version.

FrenchBen:
I'll get the toolbox version bumped - I was hoping a couple of pending PR would be merged prior to the inclusion into Toolbox.

gnachman:
Thanks, @FrenchBen! Much appreciated :)

PR #874, opened 2015-07-28
URGENT: Fixed iso and version issues

FrenchBen:
It looks like docker-machine recently changed their iso naming convention from `boot2docker.iso` to `boot2docker-virtualbox.iso` https://github.com/docker/machine/blob/master/drivers/virtualbox/virtualbox.go#L28

Furthermore the actual iso content versioning no-longer identifies itself as `Boot2Docker-v1.766` as it previously did, but rather contains docker version information, such as `Docker v1.7.1`

This patch changes the iso filename and version check accordingly - Furthermore the code was made a bit more flexible by passing the machine name to the methods, with an ES6 default of the machine name returned by the `name()` method.

Signed-off-by: FrenchBen <me@frenchben.com>

### DIFF
--- a/src/utils/DockerMachineUtil.js
+++ b/src/utils/DockerMachineUtil.js
@@ -14,10 +14,10 @@ var DockerMachine = {
   name: function () {
     return NAME;
   },
-  isoversion: function () {
+  isoversion: function (machineName = this.name()) {
     try {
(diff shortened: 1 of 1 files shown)


jeffdm:
Merging, thanks!

PR #926, opened 2015-08-13
Use new forgot password link

bithavoc:
Fixes `404 Not Found` when trying to use the Forgot Password link.


Elesant:
Thank you so much for submitting this PR! This is fixed in #935

PR #935, opened 2015-08-14
Bug: Incorrect Reset Password URL

scalopus:
Incorrect Forget Password Link


Elesant:
Ah, didn't know Hub 2.0 uses a different URL. Will fix it now. Thanks for the issue!

Elesant:
LGTM!

PR #983, opened 2015-08-27
Updated Container restart

FrenchBen:
Updates the DockerUtil restart functionality so that it follows the Start/Stop methods more closely and prevents loss of port mapping


jeffdm:
LGTM

PR #992, opened 2015-08-28
Simple kitematic cfg support for VM

FrenchBen:
This is a quick mod to allow 'advance' users to point Kitematic to their VM and if the VM doesn't exist, create it with the optional memory size.

Filename and location: `~/.kitematicfg`

Expected format:
```json
{
    "virtualbox": {
        "name": "default",
        "memory": "2048"
    }
}
```

### DIFF
--- a/src/utils/DockerMachineUtil.js
+++ b/src/utils/DockerMachineUtil.js
@@ -10,7 +10,14 @@ var DockerMachine = {
     return resources.dockerMachine();
   },
   name: function () {
-    return 'default';
+    var vmname;
+    var kitematicfg = util.kitematicfg();
+    if (kitematicfg.virtualbox && kitematicfg.virtualbox.name) {
+      vmname = kitematicfg.virtualbox.name;
+    } else {
+      vmname = 'default';
+    }
+    return vmname;
   },
   isoversion: function (machineName = this.name()) {
     try {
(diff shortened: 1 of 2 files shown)


FrenchBen:
@mchiang0610 @jeffdm Could you QA this change and see if it's applicable to the project?

Thanks!

jeffdm:
@FrenchBen sorry for the delay. Thanks for the PR! My initial feeling is that we shouldn't introduce a new config file. Most users will probably create their VM before seeing this feature - meaning they need to delete their VM and re-create it for this to work.

FrenchBen:
agreed - I did make some changes to it to be more of a VM selector in another PR

bwidmeier:
I would just like to voice my support for this feature. I agree that it has limited discoverability, and even that users may discover it after beginning to build out the 'default' machine, but given that this is more targeted at advanced users anyways, I'm not sure that that matters too much. If a user wants to go through that process, why stop them?

I should note that actually integrating multi machine support into the UI would be a better option, but this config file option seems easy to do and helpful in the meantime.

FrenchBen:
@bwidmeier I'm working on adding proper machine support into the UI, which prevents the need for this conf file :)

bwidmeier:
@FrenchBen Thanks a lot! Is there an issue that I could subscribe to?

FrenchBen:
@bwidmeier Checkout https://github.com/kitematic/kitematic/issues/1061

Issue #746, opened 2015-07-01
mongo:latest crashes on restart


vicary:
It runs well on the first download, but when I stop it never start again.

Here is the last lines of stdout.

```
2015-07-01T12:50:15.888+0000 W -        [initandlisten] Detected unclean shutdown - /data/db/mongod.lock is not empty.
2015-07-01T12:50:15.897+0000 I STORAGE  [initandlisten] ************** 
old lock file: /data/db/mongod.lock.  probably means unclean shutdown,
but there are no journal files to recover.
this is likely human error or filesystem corruption.
please make sure that your journal directory is mounted.
found 1 dbs.
see: http://dochub.mongodb.org/core/repair for more information
*************
2015-07-01T12:50:15.904+0000 I STORAGE  [initandlisten] exception in initAndListen: 12596 old lock file, terminating
2015-07-01T12:50:15.904+0000 I CONTROL  [initandlisten] dbexit:  rc: 100
```

Sorry I'm quite new to docker, while you guys are fixing this, can I like ssh in and delete the lock file myself? How can I do so?

TeckniX:
@vicary you can do the following to attach to the container and do some clean-up:
```$ docker-machine ssh dev ```
(will log you in to the actually VM)
```bash
docker@dev:~$ docker start -ai <container-id>
```
simply replace the container-id above with your monogodb container id obtained via `docker ps -a`
That should provide you with shell access to clean-up the data and then start the container.

jeffdm:
Thanks for the issue, are volumes enabled?

vicary:
@TeckniX Thanks!

@jeffdm Volumes are not mapped to Mac folders. Interestingly enough, it restarts normally in another Mac. Not sure but I will try again and update here.

vicary:
Funny enough, I've got the steps to fix this:

1. Mount the /data/db volume to Mac
2. Remove it from setting
3. Docker will start by kitematic automatically
4. Mounted folder is empty so it can be started
5. `docker exec mongo rm /data/db/mongod.lock`

Then it can be restarted normally.

Since this is a Macbook Air I suspect it is related to the Mac sleep affecting VM state and ultimately killing the mongod process without proper shutdown.

virgafox:
[...]

PR #1000, opened 2015-08-31
Fixed start/stop issues

FrenchBen:
This fixes the issue seen in many daemons such as https://github.com/kitematic/kitematic/issues/746 - By default the docker CLI uses a timeout of 10s before killing the process, Kitematic did not, which caused issues when a container had a cleanup method on shutdown.

This patch adds the same 10seconds wait time before killing the container.

### DIFF
--- a/src/utils/DockerUtil.js
+++ b/src/utils/DockerUtil.js
@@ -262,7 +262,7 @@ export default {
   },
 
   restart (name) {
-    this.client.getContainer(name).stop(stopError => {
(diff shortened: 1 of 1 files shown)


jeffdm:
Woo! Issue #1000 

Looking

jeffdm:
Tested this out locally, works great! Thanks Ben!

FrenchBen:
Sweet!

PR #1020, opened 2015-09-08
Added V2 search and pagination

FrenchBen:
Updated search to use the V2 standard (same as hub.docker.com) and added pagination for navigation.
Will probably add the same sorting functionality as found on hub also. (hence the comments)

Signed-off-by: FrenchBen <me+git@frenchben.com>

### DIFF
--- a/src/utils/RegHubUtil.js
+++ b/src/utils/RegHubUtil.js
@@ -8,6 +8,7 @@ import tagServerActions from '../actions/TagServerActions';
 
 let REGHUB2_ENDPOINT = process.env.REGHUB2_ENDPOINT || 'https://hub.docker.com/v2';
 let searchReq = null;
+let PAGING = 24;
 
 module.exports = {
   // Normalizes results from search to v2 repository results
@@ -24,7 +25,7 @@ module.exports = {
     return obj;
   },
 
-  search: function (query, page) {
+  search: function (query, page, sorting = null) {
     if (searchReq) {
       searchReq.abort();
       searchReq = null;
@@ -33,10 +34,18 @@ module.exports = {
     if (!query) {
       repositoryServerActions.resultsUpdated({repos: []});
     }
+    /**
+     * Sort:
+     * All - no sorting
+     * ordering: -start_count
+     * ordering: -pull_count
+     * is_automated: 1
+     * is_official: 1
+     */
 
     searchReq = request.get({
-      url: 'https://registry.hub.docker.com/v1/search?',
-      qs: {q: query, page}
+      url: `${REGHUB2_ENDPOINT}/search/repositories/?`,
+      qs: {query: query, page: page, page_size: PAGING, sorting}
     }, (error, response, body) => {
       if (error) {
         repositoryServerActions.error({error});
@@ -44,10 +53,14 @@ module.exports = {
 
       let data = JSON.parse(body);
       let repos = _.map(data.results, result => {
+        result.name = result.repo_name;
         return this.normalize(result);
       });
+      let next = data.next;
+      let previous = data.previous;
+      let total = Math.floor(data.count / PAGING);
       if (response.statusCode === 200) {
-        repositoryServerActions.resultsUpdated({repos});
+        repositoryServerActions.resultsUpdated({repos, page, previous, next, total});
       }
     });
   },
@@ -99,7 +112,8 @@ module.exports = {
 
   tags: function (repo, callback) {
     hubUtil.request({
-      url: `${REGHUB2_ENDPOINT}/repositories/${repo}/tags`
+      url: `${REGHUB2_ENDPOINT}/repositories/${repo}/tags`,
+      qs: {page: 1, page_size: 100}
     }, (error, response, body) => {
       if (response.statusCode === 200) {
         let data = JSON.parse(body);
@@ -122,7 +136,8 @@ module.exports = {
     let namespaces = [];
     // Get Orgs for user
     hubUtil.request({
-      url: `${REGHUB2_ENDPOINT}/user/orgs/?page_size=50`
+      url: `${REGHUB2_ENDPOINT}/user/orgs/`,
+      qs: { page_size: 50 }
     }, (orgError, orgResponse, orgBody) => {
       if (orgError) {
         repositoryServerActions.error({orgError});
--- a/src/actions/RepositoryActions.js
+++ b/src/actions/RepositoryActions.js
@@ -7,9 +7,9 @@ class RepositoryActions {
     regHubUtil.recommended();
   }
 
-  search (query) {
-    this.dispatch({});
-    regHubUtil.search(query);
+  search (query, page = 1) {
+    this.dispatch({query, page});
+    regHubUtil.search(query, page);
   }
 
   repos () {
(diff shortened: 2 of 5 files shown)


jeffdm:
Awesome! Looking!

jeffdm:
Thanks Ben! Two UX notes:

1. Should the page switcher be glued to the bottom?
2. The page selector seems to disappear temporarily when I click on a page and it starts loading. Is there an easy way to make it stay visible?

FrenchBen:
Not sure UX wise which is better, perhaps @Elesant can chime in?

The page selector does get updated with every call - The idea was to mimic the look and feel of https://hub.docker.com/search/?q=node. I'm open to suggestions for changes

jeffdm:
@Elesant ping! Can you take this for a spin?

Elesant:
Thanks @FrenchBen. This is looking good! 

We should improve upon Hub and make an even better image browsing experience on Kitematic to improve the experience of using Docker overall.

My suggestions on the UX:

1.Infinite scroll is preferred more than page selectors. Since our users are "browsing" images, we'd want them to have a similar experience to most websites such as Pinterest, Facebook etc.

2. If infinite scroll is challenging or has performance issues, we can fallback to page selector. I think the page selector component can simply stay at the bottom of the page rather than fixed at the bottom. When users scroll down, they can already see that there are more pages.

3. There is no reason to limit the number of pages shown to only 3 (we can bump up the number or use a pattern similar to Google's page selector). One advantage to having page selector is quick access to memorized page numbers.

4. The page selector no longer disappears when the page is loading on Hub. If we are using page selector, Kitematic will feel much more smooth and slick if the selector doesn't appear and reappear when pages are loading.

FrenchBen:
Infinite scroll is a possibility, but currently would break due to performance issue - Until that part is revamped we'll need to keep the paging solution. 
I'll work on updating the paging to show up to 10 or add a dropdown for quick page selection? I'll see how I can make it all work smoothly and give Kitematic the edge.

jeffdm:
I wouldn't mind merging this interim once the page viewer doesn't disappear.

FrenchBen:
I'll try to knock this out tomorrow.

FrenchBen:
@jeffdm I've made the updates as shown by @Elesant 
Currently the paging only applies to the results from hub search. Recommended/Private repo do not get paging, which is something we'll probably have to discuss/figure out. 

I also found 1 odd filter case which prevented certain results from showing up in Kitematic even though they were returned by the Hub api. I'll ping you about this to see how we want to tackle it.

thaJeztah:
Looks like these commits are not signed off properly

FrenchBen:
@thaJeztah I signed the first one as I thought it to be the one that mattered the most. I can amend the last commit and sign it if you'd like?

thaJeztah:
@FrenchBen basically, *each* commit should be signed off. Also, the sign-off should container your real name, not your GitHub handle; see https://github.com/docker/docker/blob/master/CONTRIBUTING.md#sign-your-work

In the main [Docker repository](https://github.com/docker/docker) we also usually ask contributors to "squash" their commits to logical groups. Most of the time this means that a PR contains only contains a single commit. The docker documentations has a contribution guide that describes this process; see for example this page: https://docs.docker.com/project/work-issue/

note: I'm not a maintainer in this repository, but most repositories in the Docker project follow the same rules here.

Hope this helps! Thanks a lot for your contributions, it's really appreciated :+1:

FrenchBen:
:+1: makes sense - I'll go ahead and clean all of this up. I should've known better. 

Regarding the email address I've been skeptical about it, due to possible email harvesting of 'Signed-off-by:' part - it'd be nice if you could register your email and obtain a hash to sign your commit.

thaJeztah:
@FrenchBen I understand your concern; I'm using a separate e-mail address for my sign-off's but overall, I get very little spam. Interesting thought though, I'll ask around if using a hash would be an option somehow.

jeffdm:
Small UI bug: page icons stay highlighted after clicking on them. For example, click on [ 2 ] and [ 3 ] stays highlighted.

screen shot 2015-10-21 at 4 18 29 pm

jeffdm:
Also: is there a possible way to make the page selector update as soon as it's clicked?

FrenchBen:
Latest push should fix the weird highlight that you previously saw - @jeffdm can you take a look?

jeffdm:
@FrenchBen looking - sorry for the delay

jeffdm:
Seems to work well now! A couple last things:

- Maybe, in the RepositoryStore, add a handler for the `search` action. Could this update the highlighted page number immediately?

- The `<<` and `>>` buttons seem to go to the first and last page. Just to check: is this intended? It kind of felt like they should go to the next and prev page. I think either is fine

src/utils/DockerMachineUtil.js
```diff
@@ -85,9 +85,9 @@ var DockerMachine = {
   regenerateCerts: function (machineName = this.name()) {
     return util.exec([this.command(), 'tls-regenerate-certs', '-f', machineName]);
   },
-  state: function (machineName = this.name()) {
-    return this.info(machineName).then(info => {
-      return info ? info.state : null;
+  status: function (machineName = this.name()) {
```
jeffdm: No big deal but this seems to be from another change - is master merged into this?

src/utils/RegHubUtil.js
```diff
@@ -8,6 +8,7 @@ import tagServerActions from '../actions/TagServerActions';
 
 let REGHUB2_ENDPOINT = process.env.REGHUB2_ENDPOINT || 'https://hub.docker.com/v2';
 let searchReq = null;
+let paging = 24;
```
jeffdm: Capitalize since it's a constant.

src/utils/RegHubUtil.js
```diff
@@ -99,7 +112,8 @@ module.exports = {
 
   tags: function (repo, callback) {
     hubUtil.request({
-      url: `${REGHUB2_ENDPOINT}/repositories/${repo}/tags`
+      url: `${REGHUB2_ENDPOINT}/repositories/${repo}/tags`,
+      qs: {'page': 1, 'page_size': 100}
```
jeffdm: Don't need to put object keys in quotes - only if they contain `-`.

i.e. you can do `{page: 1, page_size: 100}`

src/utils/RegHubUtil.js
```diff
@@ -122,7 +136,8 @@ module.exports = {
     let namespaces = [];
     // Get Orgs for user
     hubUtil.request({
-      url: `${REGHUB2_ENDPOINT}/user/orgs/?page_size=50`
+      url: `${REGHUB2_ENDPOINT}/user/orgs/`,
+      qs: { 'page_size': 50 }
```
jeffdm: Same as above - no quotes needed.

jeffdm:
A few code comments then this is ready to merge! AWESOME!

src/utils/DockerMachineUtil.js
FrenchBen: It wasn't - just now merged it. Good catch

FrenchBen:
Regarding the `<<` and  `>>` I'm open to either - I felt that going next could be done by clicking on the numbers, yet going to the first or last page wasn't as easy. Hence the current iteration. LMK what would be best UX wise

jeffdm:
LGTM! Feel free to merge!

jeffdm:
:boom: !

thaJeztah:
Thanks Ben!

PR #1026, opened 2015-09-10
Improves relative paths for react-router

saiberz:
Improves relative paths for react-router

### DIFF
--- a/src/routes.js
+++ b/src/routes.js
@@ -32,24 +32,24 @@ var App = React.createClass({
 
 var routes = (
   <Route name="app" path="/" handler={App}>
-    <Route name="account" path="/account" handler={Account}>
-      <Route name="signup" path="/account/signup" handler={AccountSignup}/>
-      <Route name="login" path="/account/login" handler={AccountLogin}/>
+    <Route name="account" path="account" handler={Account}>
+      <Route name="signup" path="signup" handler={AccountSignup}/>
+      <Route name="login" path="login" handler={AccountLogin}/>
     </Route>
-    <Route name="containers" handler={Containers}>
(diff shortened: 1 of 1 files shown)


saiberz:
captura de pantalla 2015-09-10 a las 12 48 05

jeffdm:
Thanks! Awesome!

One last thing, can you sign your commit?

```
$ git commit --amend -s --no-edit
$ git push -f
```

saiberz:
Done :)

src/routes.js
```diff
@@ -32,24 +32,24 @@ var App = React.createClass({
...
         </Route>
       </Route>
-      <Route name="new" path="containers/new">
+      <Route name="new" path="new">
         <DefaultRoute name="search" handler={NewContainerSearch}/>
-        <Route name="pull" path="containers/new/pull" handler={NewContainerPull}></Route>
+        <Route name="pull" path="pull" handler={NewContainerPull}></Route>
       </Route>
       <Route name="preferences" path="/preferences" handler={Preferences}/>
       <Route name="about" path="/about" handler={About}/>
```
FrenchBen: Shouldn't the above be cleaned up as well (no root '/' since part of the parent)?

FrenchBen:
Other than my small comment LGTM :+1:

src/routes.js
saiberz: Agree, maybe both should have their own handler as well as account has or is it right inside containers?

FrenchBen: I think removing the initial '/' from their path is good for now and matches the work done above. They already are outside of containers, which is how it should be.

saiberz: :+1:

jeffdm:
LGTM

PR #1038, opened 2015-09-14
Added support for ORG repos

FrenchBen:
This adds support for organization repos - Previously only a user's repo would be shown, not the org he belonged to.

### DIFF
--- a/src/utils/RegHubUtil.js
+++ b/src/utils/RegHubUtil.js
@@ -6,7 +6,7 @@ import hubUtil from '../utils/HubUtil';
 import repositoryServerActions from '../actions/RepositoryServerActions';
 import tagServerActions from '../actions/TagServerActions';
 
-let REGHUB2_ENDPOINT = process.env.REGHUB2_ENDPOINT || 'https://registry.hub.docker.com/v2';
+let REGHUB2_ENDPOINT = process.env.REGHUB2_ENDPOINT || 'https://hub.docker.com/v2';
 let searchReq = null;
 
 module.exports = {
@@ -73,7 +73,7 @@ module.exports = {
         }
 
         request.get({
-          url: `${REGHUB2_ENDPOINT}/repositories/${name}`,
+          url: `${REGHUB2_ENDPOINT}/repositories/${name}`
         }, (error, response, body) => {
           if (error) {
             repositoryServerActions.error({error});
@@ -104,10 +104,14 @@ module.exports = {
       if (response.statusCode === 200) {
         let data = JSON.parse(body);
         tagServerActions.tagsUpdated({repo, tags: data.results || []});
-        if (callback) { callback(null, data.results || []); }
+        if (callback) {
+          return callback(null, data.results || []);
+        }
       } else {
         repositoryServerActions.error({repo});
-        if (callback) { callback(new Error('Failed to fetch tags for repo')); }
+        if (callback) {
+          return callback(new Error('Failed to fetch tags for repo'));
+        }
       }
     });
   },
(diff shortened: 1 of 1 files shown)


src/utils/RegHubUtil.js
```diff
@@ -117,59 +121,97 @@ module.exports = {
...
-        }, (error, response, body) => {
+
+      // Get Orgs for user
+      hubUtil.request({
+        url: `${REGHUB2_ENDPOINT}/user/orgs/?page_size=50`
+      }, (orgError, orgResponse, orgBody) => {
+        if (orgError) {
+          repositoryServerActions.error({orgError});
+          if (callback) {
+            return callback(orgError);
```
jeffdm: No need to return the callback.

src/utils/RegHubUtil.js
```diff
@@ -117,59 +121,97 @@ module.exports = {
...
+
+        if (orgResponse.statusCode !== 200) {
+          let generalError = new Error('Failed to fetch repos');
+          repositoryServerActions.error({error: generalError});
+          if (callback) {
+            return callback({error: generalError});
+          }
+          return null;
+        }
+        let orgs = JSON.parse(orgBody);
```
jeffdm: Maybe put this in a try/catch block and report the error on `catch`?

src/utils/RegHubUtil.js
```diff
@@ -117,59 +121,97 @@ module.exports = {
     repositoryServerActions.reposLoading({repos: []});
 
     hubUtil.request({
-      url: `${REGHUB2_ENDPOINT}/namespaces/`,
+      url: `${REGHUB2_ENDPOINT}/namespaces/`
```
jeffdm: I don't think we need to hit `/namespaces` anymore. Instead (like below), just getting the orgs (and append the user's name) is fine.

src/utils/RegHubUtil.js
```diff
@@ -117,59 +121,97 @@ module.exports = {
...
+          let generalError = new Error('Failed to fetch repos');
+          repositoryServerActions.error({error: generalError});
+          if (callback) {
+            return callback({error: generalError});
+          }
+          return null;
+        }
+        let orgs = JSON.parse(orgBody);
+        orgs.results.map((org) => {
+          namespaces.push(org.orgname);
```
jeffdm: I think we need to append the username to this list around here.

jeffdm:
Awesome! A few comments, but overall this is something many people are asking for. Thanks for tackling it!

src/utils/RegHubUtil.js
FrenchBen: I had some errors with lint because of the inconsistent return - I'll make the changes

src/utils/RegHubUtil.js
FrenchBen: I'll update it to a try-catch; this was re-using existing code for simplification, but might as well update it.

src/utils/RegHubUtil.js
FrenchBen: To confirm - I'll remove the namespaces call, use only the orgs obtained and append the username to the list?

michaelwilde:
you guys are awesome...  is this an 0.8.5 build perhaps... coming.. today, tomorrow?

halr9000:
+1 we are looking at how best to use org repos, this helps a lot

jeffdm:
@michaelwilde  Will cut a new build first thing when this is merged!

jeffdm:
@FrenchBen this is awesome work. I've verified that it works on my end.

Code LGTM! Merging.

FrenchBen:
:+1:  thanks for merging this @jeffdm

PR #1093, opened 2015-10-08
remove non-working ACCESS URL from home page and replace it with simple IP & PORT (same as on settings page)

andreysaksonov:
rationale behind this is that most containers don't have any web UI to use as "access url". maybe I am wrong but usually I use KItematic for some databases like postgres or mongo or redis etc.

### DIFF
--- a/src/components/ContainerHomePreview.react.js
+++ b/src/components/ContainerHomePreview.react.js
@@ -69,11 +69,11 @@ var ContainerHomePreview = React.createClass({
     } else {
       var ports = _.map(_.pairs(this.props.ports), pair => {
         var key = pair[0];
-        var val = pair[1];
+        var {ip, port, url, error} = pair[1];
         return (
           <tr key={key}>
             <td>{key}</td>
-            <td>{val.display}</td>
+            <td>{ip}: {port}</td>
           </tr>
         );
       });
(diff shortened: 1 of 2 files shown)


FrenchBen:
Instead of taking it completely out, why not make 1 quick/simple change like so:
`val.display = (val.display) ? val.display : val.ip + ':' + val.port;`

andreysaksonov:
first of all {val.display} will always produce empty string after changes were made to allow user to edit port manually.

thing that was previously called "access url" is available in scope, however it is just simple combination of http://{ip}:{port}, so it will lead to something like "http://192.168.99.100:5432" which doesn't make any sense for user because there is no web server at this address. the idea of detecting web UI is user-friendly, however we need smart solution here to make it useful. i was thinking about loading this "access url" with http client and check if there some http answer or not. however this feature requires some design and analysis from UX prospective.

FrenchBen:
Exactly, being able to determine what the container is serving is difficult - We can look at port numbers and make an educated guess, but most of the time the user will know better. 

The sample code I provided was to illustrate a lighter change that may support the `val.display` should it ever not be empty. If it was added once, it's possible that the API used to respond to it and could be used in the future (aka future proof change). 

Either way I think the change is needed, even though it forces the user to copy & paste the text displayed

FrenchBen:
#closing this PR as it's out-dated but I will implement the solution in one of the existing PR that should be merged.
Thanks for sharing the code to re-use. I'll make sure to mention you in the PR that will have the changes.

PR #1098, opened 2015-10-09
Enabled scrolling on new container

FrenchBen:
Somehow we lost the ability to scroll on the New/Search Container page. This adds some overflow magic.


jeffdm:
@FrenchBen good catch! Thanks!

PR #1118, opened 2015-10-16
Updating with Hugo 15

moxiegirl:
Updates the content for use with Hugo 15 patch. @SvenDowideit created this patch
The patch supports GitHub source relative links

Signed-off-by: Mary Anthony <mary@docker.com>


mchiang0610:
@moxiegirl LGTM. The code shouldn't break anything.

PR #696, opened 2015-06-22
Linux support

zedtux:
This branch add the missing Linux support.

Basically it avoid the wizard to run and force `dockerode` to use `/var/run/docker.sock` as a `socketPath`.

Regarding the in-container-shell only the GNOME terminal has been implemented (as I'm using Debian) but a warning message asking to open an issue in the project will be shown in other cases so that we can implement them one after the other.

The missing part of this merge request is the packaging. I don't really want to deal with that and let packagers taking care of this task but I'm available to help in case of questions and Grunt task fabrication.

PR #1143, opened 2015-10-26
Linux support

FrenchBen:
There's a lot in here, but this works in vanilla Debian (after following the wiki instructions) and works properly on my Mac. 
I lack the windows environment, so if @jeffdm can take a look?

Huge thanks to @zedtux for all of the work on this - I only cleaned up a few things


FrenchBen:
#closing in favor of #696

PR #1145, opened 2015-10-26
Add stopping state to fix immediately killing containers

jeffdm:
- The docker daemon emits `kill` events when containers are stopped. We can use this to disable the start, restart and exec buttons while the container shuts down.

- Added a 5 second timeout


FrenchBen:
LGTM

FrenchBen:
Still LGTM thanks for the quick fix

jeffdm:
Thanks @FrenchBen !

PR #1153, opened 2015-10-28
Fill empty spaces

saiberz:
I'm not sure if this space's intentionally left blank

### DIFF
--- a/styles/new-container.less
+++ b/styles/new-container.less
@@ -181,7 +181,7 @@
       display: flex;
       flex: 1 0;
       position: relative;
-      min-width: 35%;
+      min-width: @image-card-width;
       height: @image-card-height;
       border-radius: @border-radius;
       background-color: white;
--- a/styles/new-container.less
+++ b/styles/new-container.less
@@ -181,20 +181,18 @@
       display: flex;
       flex: 1 0;
       position: relative;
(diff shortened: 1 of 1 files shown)


FrenchBen:
Thank you for the PR - The 'empty' space is actually a result of the window size, where 3 image boxes cannot fit and thus wrap around. 
If the window was slightly bigger, then it would allow for them to span across the entire window width. 

Unfortunately your fix wouldn't allow for more than 2 images to be displayed on 1 row, which could look very ugly on larger screens. Could you try to tweak it so that 2 images fit 100% but if more than 2 could fit, then you see more? (3, 4, etc)

saiberz:
@FrenchBen how about this? 
screen shot 2015-10-28 at 18 01 02
screen shot 2015-10-28 at 18 02 41

It fits the full width and can wrap many elements

FrenchBen:
Much better, but if there is only '1' container on a row, it is stretched to the entire width of the window.[recommended](https://cloud.githubusercontent.com/assets/3965324/10806865/bdb3251e-7d96-11e5-9fd9-bc7c31192831.png)

FrenchBen:
Thanks for the PR - Closing this for now as there isn't a proper solution.

PR #1161, opened 2015-10-30
Fixes #1061 with VM name selection

### LINKED ISSUE #1061: Feature request - Change VM name ( " default " )
Please allow to select name for the VM that Kitematic uses for dockers, and allow to rename the existent VM.

FrenchBen:

### DIFF
--- a/src/components/Header.react.js
+++ b/src/components/Header.react.js
@@ -26,7 +26,8 @@ var Header = React.createClass({
     document.addEventListener('keyup', this.handleDocumentKeyUp, false);
 
     accountStore.listen(this.update);
-
+    // remove listener if exists
+    ipc.removeAllListeners('application:update-available');
     ipc.on('application:update-available', () => {
       this.setState({
         updateAvailable: true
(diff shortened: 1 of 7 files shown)


FrenchBen:
Tried to QA with Windows but currently getting some errors from virtualbox. Will wait for new toolbox to be pushed out

jeffdm:
Taking a look!

FrenchBen:
FYI, it looks like the branch needs to be updated prior to final push

pmario:
@FrenchBen Did this PR quietly die?

FrenchBen:
@pmario it did due to machine management. The VM selection was fine, but for the most part, there was no mounting of folders and misc actions were broken.

pmario:
Yea. I found out already, that the whole system is heavily connected to vbox. From a programmers point of view. Way to heavy. ...

IMO kitematic needs a plugin mechanism, where vbox handling is a plugin too. 

This would give us (the community with different requirements) the possibility to remove the vbox battery and replace it. 

At the moment I'm investigating to create shared folders with hyper-v and using container tech and not vbox tech.

FrenchBen:
I agree with that approach, unfortunately the current heavy vbox integration makes it difficult. 
We've considered re-writing the app completely with the  plugin in mind, but there's a lot to do to make that happen. 
I'm open to suggestions if you see an easy way to get that working?

PR #1169, opened 2015-11-03
Better docker-machine error reports

jeffdm:


src/utils/DockerMachineUtil.js
```diff
@@ -24,11 +24,11 @@ var DockerMachine = {
   version: function () {
     return util.exec([this.command(), '-v']).then(stdout => {
       try {
-        let tokens = stdout.split(' ');
-        if (tokens.length < 3) {
-          return Promise.resolve(null);
+        var match = stdout.match(/(\d+\.\d+\.\d+).*/);
```
FrenchBen: Only change I would make is rename the var to something that doesn't match the method name:
`var matched`
vs 
`var match`

src/utils/DockerMachineUtil.js
```diff
@@ -157,6 +157,14 @@ var DockerMachine = {
         util.exec([path.join(process.env.RESOURCES_PATH, 'terminal'), `DOCKER_HOST=${machineUrl} DOCKER_CERT_PATH=${path.join(util.home(), '.docker/machine/machines/' + machineName)} DOCKER_TLS_VERIFY=1 ${cmd}`]).then(() => {});
       });
     }
+  },
+  virtualBoxLogs: function (machineName = this.name()) {
+    let logsPath = path.join(util.home(), '.docker', 'machine', 'machines', machineName, machineName, 'Logs', 'VBox.log');
+    if (fs.existsSync(logsPath)) {
```
FrenchBen: I think npm recommends doing a try/catch on readFileSync instead of doing an exist check.
```
let logData = null;
try {
  logData = fs.readFileSync(logsPath, 'utf8');
} catch (e) {
  console.error(e);
}
return logData;
```

jeffdm:
Thanks @FrenchBen! Updated

FrenchBen:
LGTM!

PR #1198, opened 2015-11-10
Increasing width by 70px to have 3 columns of the AppStore.

mchiang0610:
Increasing width by 70px to have 3 columns of the AppStore.


jeffdm:
LGTM

PR #1199, opened 2015-11-10
Improved log performance and reliability

jeffdm:
For performance reasons, only allow attaching to one `active` container at a time. This container is set to the one being viewed by the user.


mchiang0610:
LGTM. Works very well on Windows!

src/stores/ContainerStore.js
```diff
@@ -158,6 +158,35 @@ class ContainerStore {
...
+  logs ({name, logs}) {
+    let container = this.containers[name];
+
+    if (!container) {
+      return;
+    }
+
+    container.Logs = logs.split('\n');
+
+    console.log(container.Logs);
```
FrenchBen: Can we remove this debug?

src/utils/DockerUtil.js
```diff
@@ -341,16 +422,24 @@ export default {
       stream.on('data', json => {
         let data = JSON.parse(json);
 
-        if (data.status === 'pull' || data.status === 'untag' || data.status === 'delete' || data.status === 'attach') {
+        console.log('Event from daemon: ' + data.status + ' ' + data.id);
```
FrenchBen: Do we want to keep this debug/log ?

src/utils/DockerUtil.js
```diff
@@ -330,6 +332,85 @@ export default {
...
+      logStream.setEncoding('utf8');
+      logStream.on('data', chunk => logs += chunk);
+      logStream.on('end', () => {
+        containerServerActions.logs({name: this.activeContainerName, logs});
+        this.attach();
+      });
+    });
+  },
+
+  attach () {
```
FrenchBen: Seems that attach() and logs() do very similar things, could they be merged into 1 method?

FrenchBen:
A few comments but otherwise looks awesome!

src/utils/DockerUtil.js
jeffdm: They are similar in that they both call the /logs api endpoint, but they behave differently - one streams changes and the other hooks into the logs. I tried merging them in various ways, but kept leaning towards two functions.

FrenchBen:
LGTM

FrenchBen:
:+1: All looks good

Issue #1192, opened 2015-11-07
Question: Why volumes only can be specified if they are under the Users directory (Windows)


alexandrev:
I'm using Kitematic for Windows, and when I try to specify the host path for a VOLUME sentence in my docker container, I'm getting this error:

![image](https://cloud.githubusercontent.com/assets/947041/11016767/98c686e6-858a-11e5-9827-8e0327a10294.png)

If I specify a folder inside my User folder, everything is OK. But I don't understand that kind of restriction. Even more, in the Windows system, when I can have different volumes mounted with different folders and it has no sense to be restricted to your User folder.

Why is that reason and could be change for next versions at least in Windows systems? Maybe if you have to be sure that the user has full access, we can restrict the Owner of the folder or something like that...

What do you think? 

Thanks!!

mchiang0610:
@alexandrev Thanks for reporting this. We will pass this on to the Docker machine team for a fix. 
cc/ @nathanleclaire

alexandrev:
Thanks @mchiang0610, If I could do more, please let me know. I'm here to help in everything I can.

mchiang0610:
@alexandrev Can you provide us logs / your windows version as well for us to reproduce this?

alexandrev:
The windows version is windows 10 build 10586. Please, tell me which logs
you want and I'll provide it to you.

Thanks,

FrenchBen:
@alexandrev The reason behind this is linked to the VM - When the Virtualbox is created it mounts the folders within your `Users` directory, which makes it the only folder available to the containers.
You're getting the error due to this restriction and you were probably trying to mount folders outside of the `Users` path

alexandrev:
I know the error, as you said is because i'm trying to mount a volume from
another folder but why is the reason for this restriction?

Could I personalize the folder that it is mounted inside the vm?? Any way
to customize these options?

FrenchBen:
@alexandrev In theory you could add more folders to the VM - It's impossible for us to know what folders are available or unavailable, but the `Users` directory will always be there. There isn't a straight-forward way for us to check what folders have been mounted by the user within the VM; We added the check within the code to make sure that we could read/write to the folder.

Most developers will work out of their user folders/desktop and thus never see this restriction. Feel free to create a PR that allows for greater flexibility :)

alexandrev:
OK, I want to try it. We can close this question because now I know the
reason of the message.

Thank you all for the time you spend answering this question :)

[...]

PR #722, opened 2015-06-25
Allow mounting custom volume directories

zedtux:
This pull request solve the issue #376.

This pull request updates the "Volumes" sub tab panel of the "Settings" container tab with a line to add new volumes:

screenshot from 2015-06-25 20 51 02

Writing a path, clicking the plus and then 'Save' will update the container so that a new folder is available but empty.

screenshot from 2015-06-25 20 53 15

Then selecting a local folder using the 'CHANGE' button will now refresh correctly the UI (before it was refresh only when going somewhere else and then back to the tab) and also will Bind the local path to the Docker path.

screenshot from 2015-06-25 20 55 00

Now you can click the 'REMOVE' button to unlink the local folder (not new) or you can use the X button to remove completely the volume, which will re-create immediately the container (also there were a bug here with the `Binds` attribute of the Docker server API. It is now embedded in the `HostConfig` attribute while Kitematic was setting the `Binds` attribute at the root of the JSON).

screenshot from 2015-06-25 20 57 33

Last but not least, the 'RESET' button allow the user to revert the changes until he clicks the 'SAVE' button.
So let say the user add some volumes and use the 'CHANGE' button to link the local folder to the docker folder (clicking the 'EXEC' icon and list the folders in the terminal will show the new folders) but then he wants to rollback everything as it was before, he can click the 'RESET' button which will unlink and remove the folders automatically.

PR #1202, opened 2015-11-10
Flexible Local Folder for Volumes

alexandrev:
This change allows to define Volumes where the local folder is not limited to the Users directory but to any folder mounted in the VM.

It recovers from the VBoxManage the mounted folders and it checks the final folder agains this values, and if it matches it allows to specify the path.

This code is related to Issue #1192 opened as a question.

Please, if you have any question, only let me know.


FrenchBen:
:+1: thanks for creating this PR - I'll check it out tomorrow, but at first glance it seems that you found a good way to check for shared folders. 

Can you confirm that this also compiles properly on Windows?

alexandrev:
Yeah, I'm using Windows as dev environment so I only could check it works
properly on Windows platform.

My current version is W10 Build 10586.

I'm sure we can optimize the code but I think it's a good first approach to
get it done.

Anything you can reach me here :)

src/components/ContainerSettingsVolumes.react.js
```diff
@@ -15,12 +15,21 @@ var ContainerSettingsVolumes = React.createClass({
       }
 
       var directory = filenames[0];
-
-      if (!directory || directory.indexOf(util.home()) === -1) {
+      var founded = false;
+        if(directory){
```
FrenchBen: Can you fix the indentation and check that spacing around keywords 'if' are respected?

src/components/ContainerSettingsVolumes.react.js
```diff
@@ -82,16 +91,16 @@ var ContainerSettingsVolumes = React.createClass({
     var homeDir = util.isWindows() ? util.windowsToLinuxPath(util.home()) : util.home();
     var mounts= _.map(this.props.container.Mounts, (m, i) => {
       let source = m.Source, destination = m.Destination;
-      if (!m.Source || m.Source.indexOf(homeDir) === -1) {
+      /*if (!m.Source || m.Source.indexOf(homeDir) === -1) {
```
FrenchBen: Don't comment out code, simply take it out - We have git versioning to roll back if needed

alexandrev:
Hi @FrenchBen, Sorry for the things you point out. You are absolutely right. My bad! I think I push all the changes you requested. Please, take a look if now is OK for you and sorry again.

src/utils/SetupUtil.js
```diff
@@ -82,6 +82,9 @@ export default {
           machineVersion
         });
 
+        virtualBox.getShareDir(machine.name());
```
FrenchBen: If a user adds a shared directory, it will only be seen until the next kitematic restart - Is this the correct approach?

src/utils/Util.js
```diff
@@ -8,6 +8,7 @@ import remote from 'remote';
 var app = remote.require('app');
 
 module.exports = {
+  folders: [],
```
FrenchBen: Is this used anywhere?

src/utils/VirtualBoxUtil.js
```diff
@@ -59,6 +59,26 @@ var VirtualBox = {
...
+        if(items[i].startsWith("Name")){
+          elements.push(myRegexp.exec(items[i])[1]);
+        }
+      }
+      util.folders = elements;
+      return true;
+    }).catch((err) => {
+      return false;
+    });
+
```
FrenchBen: Same as above, make sure this matches ESLint specifics (spacing, indents, etc). 
If you're using Atom.io to do your development, you can install the 'linter' and 'linter-eslint' packages. The `.eslintrc` file should then do the rest.

alexandrev: OK, I'm using Atom so I'm going to install the plugins you recommend and fix these problems.

src/utils/Util.js
alexandrev: Yeah, the folder attribute is where you stored the array with the mounted folders in the VM. You use this attribute in rc/components/ContainerSettingsVolumes.react.js line 20, and src/utils/VirtualBoxUtil.js line 76.

src/utils/SetupUtil.js
alexandrev: What you are comment is totally right. I think that the operation of add a new folder to the machine is not a "normal" activity when you are working, it is more usual an activity you do so few times when you are configuring your enviroment. So, I think recovering once when the program starts is enough, but if you think it's better another approach we can do this call when anyone is trying to change the volume.. 

I'm open to all your considerations, you are the master here :)

src/components/ContainerSettingsVolumes.react.js
```diff
@@ -82,16 +91,11 @@ var ContainerSettingsVolumes = React.createClass({
     var homeDir = util.isWindows() ? util.windowsToLinuxPath(util.home()) : util.home();
```
FrenchBen: this var can be removed since you no longer check for the home dir.

src/utils/SetupUtil.js
FrenchBen: haha no such thing as the master, just an observation. I think you're right, the user would add more folders and restart or start kitematic after making the changes. 
When in doubt restart, right?

src/utils/Util.js
FrenchBen: If this is an array of mounted folders within a VM, shouldn't that be a VirtualBoxUtil.js variable instead? `this.folders = elements;` ?

FrenchBen:
Doing some QA on the code you currently have I'm having a couple of issues while using the hello-world-nginx

- After adding a Volume, trying to remove it, I get the following JS error:
`ContainerSettingsVolumes.react.js:132 Uncaught TypeError: Cannot read property 'replace' of null`

Also mounting a volume on my Desktop (within the user dir), removing it and closing kitematic and opening it again, I get the following within your repo:

bad-volume

I confirmed this doesn't happen on Master.

alexandrev:
Hi @FrenchBen , sorry to hear that, I'm going to try to reproduce it here, and see which is the root cause. I'll get back to you.

alexandrev:
Hi @FrenchBen , I think I fix the bug you pointed out. The problem was I removed the "No Folder" condition and when the source was null, it crashes! Sorry! My bad! 

please, check it now and if you have any problem, I could fix it right away!

alexandrev:
Any more feedback on this PR? Thx!

FrenchBen:
Wont' be able to look at it till monday - meanwhile, can you squash your commits into logical commits, instead of the multiple "fix this" "removed that"?

alexandrev:
Of course I will do. Thanks for everything

kotowick:
Excited for this feature! Anyone have an ETA?

alexandrev:
Hi @FrenchBen . I have squashed the commits here as well to fix the same thing that you reported in the other PR . I hope with this change everying is OK. 

Regarding to the @kotowick we are doing some QA to see that there is no bugs in the PR before any kind of merge. Regarding to the merge I have no clue about that but I'd like to be soon!!! :+1:

kotowick:
Thanks @alexandrev, looking forward to this feature :+1:

kotowick:
@alexandrev @FrenchBen hoping to have this in the next release?

FrenchBen:
I mounted a folder from the tmp dir in VirtualBox and made it 'automount' + 'permanent'
folder


I then opened Kitematic based on your PR, created a container based on the `kitematic/hello-world-nginx` and tried to mount a folder outside of the mounted volume and got the proper error :+1: 
When I tried to mount the volume in tmp, everything went well and it reported the volume as mounted. 

The problem was that the 'index.html' file that normally gets created/copied into the volume mounted, didn't get copied. 

Are you able to replicate this issue? If so, do you know why this is happening?

alexandrev:
I'm going to try to reproduce it, and I'll come back to you with some answers :)

alexandrev:
Hi @FrenchBen,

I'm going to describe my steps to reproduce the issue you are facing. Please, let me know if I'm doing something wrong to do it. 

First, I add a new local folder inside the Docker VM machine, as you can see here:

image

Then, I specify the mapping in the Volume section as you can see here:

image

I start the container as you can see here:

image

And I can found the index.html file inside the DockerVM file:

image

Is everything OK? Or am I missing something? 

Thanks!

FrenchBen:
Weird on my machine, going to `D:\Temporal` there was no file created - What permissions did you give the folder when you mounted it via VirtualBox?

alexandrev:
I don 't do anything   special to get it mounted. I even not mount the
folder myself but I let the boot2docker machine do it for me when I restart
the container after changing the volume settings to the new path.

FrenchBen:
My question was how you add the folder to VirtualBox?

alexandrev:
Here you are, my virtual box shared folder configuration:

image

kotowick:
@alexandrev @FrenchBen Thank you for your dedication to this feature. Let me know if I can assist.

FrenchBen:
@kotowick feel free to also do some QA on this PR and see if it works for you. The more test cases, the better

alexandrev:
I've updated the branch (I have to change a few things :) ) but I think now it's everything working as it was before the multiple updates in the master branch! 

Please, any feedback would be helpful to eventually merged this PR :)

FrenchBen:
I'm currently running into some networking issues which prevent this PR from being properly QA'd - I'll try to get those fixed to give this a proper test. 
Thank you for your patience.

bitsrfr:
I'd love to see this functionality added!

alexandrev:
Updated the branch to the latest changes on master. Also cleaned the history to do it better to the maintainer's work. I'd love to see this changes included on the branch version, so please, any help I could provide to make it happened as soon as possible let me know. Regarding the comments we can see in the conv, this feature is truly needed and appreciate it.

alexandrev:
No news?

FrenchBen:
Looks like some files are missing - I'm getting the following error when I try to mount a folder that has been setup as 'shared' in VirtualBox:
vbox

alexandrev:
Please @FrenchBen try again. Some changes got lost after the squash commits. Everything should work right now.

tiptronic:
Do we need to build it ourselves, or will there be an update to Kitematic (containing this change) anytime (very) soon ? ;)

Btw: Dare I suggest putting a changelog (or at least a version-number) on the https://www.docker.com/docker-toolbox page?

FrenchBen:
@tiptronic We're going through QA right now - Once it's pushed into the main repo, we'll make a release a push it into the Toolbox. There are a couple of pending fixes we'd like to see incorporated as well :)

tiptronic:
@FrenchBen Excellent! Great to hear! And thanks for alle the cool stuff! I love it!

tiptronic:
...and I appreciate your work on the project very much!

kotowick:
@FrenchBen excited for this! (ETA for release?)

@alexandrev thank you for the hard work :)

kotowick:
@FrenchBen  @alexandrev 

Having difficulties with mounting the volume..

Need to mount /Volumes/Development (local on OSX):

screen shot 2016-01-21 at 2 20 52 pm

After a restart on docker-machine, the volume isn't mounted.

screen shot 2016-01-21 at 2 22 14 pm

In Kitematic, I can set the "local" volume to /Volumes/Development/data to /data. Attach to the VM and nothing shows under /data.

screen shot 2016-01-21 at 2 27 44 pm

Any clues?

alexandrev:
Hi Spencer,
 
You should mount the folder inside your Linux Machine (the machine you are working inside your VirtualBox)
 
If the folder is not mounted itself you should type something like this:
 
```
mkdir -p /Volumes/Development
sudo mount -t vboxsf  Development  /Volumes/Development
```
 
Please, keep me posted abour the issues you are facing.
 
Thx

kotowick:
@alexandrev with VirtualBox (docker-machine VM) it auto mounts /Users to Users

I'm assuming it won't mount /Volumes/Development to Volumes/Development because the folder path doesn't exist on the VM?

Each time I start up VirtualBox, I have to mount the folder manually. Would be nice if all of this was automatic (I wrote a small bash script to start the VM, ssh in, create the directories, mount the volume, and then open the Kitematic app).

FrenchBen:
@kotowick I'm surprised that your screenshot doesn't auto-mount the volume:


Although I had similar inconsistent issues with this.

kotowick:
@FrenchBen surprised too. Not sure what is up...on latest release of Virtual Box. I thought that it wouldn't mount because the VM doesn't contain the local folder "/Development".

Also, lets say that I mounted /Volumes/Development to /Development on the Virtual Box VM. 
In a docker instance in Kitematic, If I mount /data to /Volumes/Development/data, it doesn't work because in the VB VM it is mounted to /Development. 

I had to change the VB VM to mount /Volumes/Development to Volumes/Development for this to all work..

alexandrev:
@kotowick as @FrenchBen said, it is a surprise because on my environment (Windows) I don't have this issue, but @FrenchBen had it sometimes as he said, but that's an "issue" that belong to VirtualBox or at least the boot2docker image inside VirtualBox. I don't think we can do more from Kitematic as we are acting like a VirtualBox client.

Anyone has the same problem? or doesn't have it? Any other QA?

FrenchBen:
@alexandrev thanks for looking into it - Adding a feature that doesn't work on the Mac (almost a windows-only thing) becomes an issue. 
It's interesting that it works on Windows, as the boot2docker specifically makes User changes manually. 

See:
https://github.com/docker/boot2docker/blob/master/Dockerfile.virtualbox#L24

alexandrev:
I dont understand your comment. I don't see any issue here. It is a
configuration related to the oracle virtualbox and boot2docker images but
not a kitematic issue. I don't think the solution is to avoid features
because you need some configuration.

But I'm only a developer who helps in my spare time, so it's your call. If
you want,  close the PR.

Have a good day

kotowick:
@FrenchBen @alexandrev is right, it is more of a VirtualBox issue. VirtualBox is not auto-mounting shared folders/drives (even if the option is set to). 

Let's keep this PR open - the feature is still needed.

FrenchBen:
@kotowick couldn't agree more! :+1:

tiptronic:
Just for my understanding: This feature doesn't make it into the Kitematic UI? (At least after updating to 1.10 I don't see any change).

alexandrev:
It's not merged yet. Sorry Andy :(

tiptronic:
Thank for the info Alexandre! Do you have any idea when that will happen?

alexandrev:
We are on QA phase, but the manpower of the core mainteners is limited so I
don't know any ETA.

tiptronic:
If you need any help (for testing) - just let me know... Unfortunately this seems to be my only (limited) way to assist...

FrenchBen:
@alexandrev @tiptronic The issue is no longer man power, but the fact that the code would only work for 1/2 of the solution.

As mentioned before, there are still some limitations to getting this to work properly:

1. Adding the folder entry in Virtualbox is a very manual process, where the options need to be set properly
2. VirtualBox is not auto-mounting shared folders/drives (even if the option is set to).

Until the above are resolved we will not be able to move forward.

tiptronic:
@FrenchBen Maybe there's slight misunderstanding... I didn't mean auto-mounting volumes on my host machine, but allowing me to add a mount-point in Kitematic, which points to a volume or folder on my Mac. On my Synology-NAS I can add a share as volume/mount-point to a docker image from the graphical UI. That's what I miss in Kitematic most.

extra-0:
For me even manually mounting /Volumes/Development under /Volumes/Development didn't work on Mac. Solved by mounting /Volumes to /Volumes with proper rights. Something like this:

    mount -t vboxsf -o gid=50,uid=1000 Volumes /Volumes

Would be nice to have a script inside the VM that would scan for the shares and add them automatically.

alexandrev:
Hi extra-0,

I open an issue to boot2docker to clarify this different behavior depending
on the OS and to address this problematic.

I hope we can find a good solution for all our needs. :)

extra-0:
So I fooled around with VBoxManage / VBoxControl. Here's what I got so far.
Mounting all shares within guest box can be done with something like this:

```
VBoxControl --nologo sharedfolder list | grep "-" | awk '{print $3}' | xargs -r mkdir {} && mount -t vboxsf -o gid=50,uid=1000 {} /{}
```

This works almost ok, except for 2 problems.

1. it also creates a "{}" folder. If someone can fix that, it would be great.
1. it doesn't handle nested folders.

Then I tried running this from host machine, but stumbled upon following:

```
VBoxManage guestcontrol {599eb0ba-9800-428f-9bf3-6d05b92e0d2e} run "which awk"
VBoxManage: error: The specified user was not able to logon on guest
VBoxManage: error: Details: code VBOX_E_IPRT_ERROR (0x80bb0005), component GuestSessionWrap, interface IGuestSession, callee nsISupports
VBoxManage: error: Context: "WaitForArray(ComSafeArrayAsInParam(aSessionWaitFlags), 30 * 1000, &enmWaitResult)" at line 938 of file VBoxManageGuestCtrl.cpp
```

If someone could help me fix these I think I would be able to make a bash script to fix automount.
It is also possible to get list of all shares using VBoxManage, but it doesn't change anything that much.

alexandrev:
Hi @extra-0 . A few questions here:

The command:

```
VBoxControl --nologo sharedfolder list | grep "-" | awk '{print $3}' | xargs -r mkdir {} && mount -t vboxsf -o gid=50,uid=1000 {} /{}
```
Are you running it inside the Boot2docker machine on the Virtualbox? 

I think if this works handling the nested folder should be no problem using (mkdir -p to create all the path and trying to mount it)

I think that's a good approach we should focus on a PR for the boot2docker project :).

I'd like to help, so please, keep me in the loop and we can work together on this!

Thx!

extra-0:
> Are you running it inside the Boot2docker machine on the Virtualbox?

Yes, however I'm not yet sure the mount works properly. `{}` folder prevents from further testing

> mkdir -p

live and learn :)

Any idea on how to run this within guest machine from the host?

alexandrev:
I don't think we can exec that from the host like a command or order. I think we should try to include inside the boot2docker machine. And open a PR to be included at the core of the Docker VM.

What do you think about that @FrenchBen @jeffdm @mchiang0610 ? Do you think this could be a good approach?

FrenchBen:
@alexandrev Yup I think the issue could be fixed from boot2docker, so that it more dynamically reads the volumes that Vbox has declared as mounted.

alexandrev:
Ok, so we are going to address the issue and the future PR to them. And after that, we could advance with this PR and everyone is winning :)

extra-0:
So more fooling around and here's what I got.

Run from host computer:
`for i in $(VBoxManage list vms | grep "\"default\"" | awk '{print $2}' | xargs VBoxManage showvminfo --details --machinereadable | grep SharedFolderNameMachineMapping | awk -F '"' '{print $2}'); do docker-machine ssh default "sudo mkdir -p /$i; sudo mount -t vboxsf -o gid=50,uid=1000 $i /$i";done`

How this works.

`VBoxManage list vms` - get list of vms in host

`grep "\"default\"" | awk '{print $2}'` - get uuid of "default"

`xargs VBoxManage showvminfo --details --machinereadable` - send this uuid to `VBoxManage showvminfo` to get params including information about shared folders

`grep SharedFolderNameMachineMapping` - get names of shares. They should correspond to the folder paths.
Otherwise one can use `SharedFolderPathMachineMapping` to get the paths and use them

`awk -F '"' '{print $2}'` - use awk to strip strings

`for i in $(...) do ... ; done` - iterate through results

`docker-machine ssh default` - connect to docker guest using `docker-machine` from docker toolbox

`sudo mkdir -p /$i` - since `docker-machine` connects us as `docker` user, go sudo and create mountpoints. `$i` is a parameter from iteration

`sudo mount -t vboxsf -o gid=50,uid=1000 $i /$i` - mount shares named $i at /$i as docker:staff

And it seems to work.
This can also be done from within guest with some minor changes, i.e. iterate over results of `VBoxControl --nologo sharedfolder list | grep "-" | awk '{print $3}'` and perform `mkdir` and `mount` directly within guest.

alexandrev:
Thx @extra-0. Great job!. Could you work on the command from guest. So we can open a PR to boot2docker, because this is feature that is good for anyone using the Docker Toolbox not only the Kitematic users. If they don't agree to merge it, we could try the other approach. Everyone agree?

extra-0:
Sure, it's even more simple:
`for i in $(sudo VBoxControl --nologo sharedfolder list | grep "-" | awk '{print $3}'); do sudo mkdir -p /$i; sudo mount -t vboxsf -o gid=50,uid=1000 $i /$i; done`

I'll submit a pull-request first thing tomorrow and post a link here.

alexandrev:
Great crack! Please, keep us posted!

--
Alexandre Vazquez

Enviado desde dispositivo móvil.

FrenchBen:
I do like this script a lot - Seems pretty simple and to the point. Thanks @extra-0 for the work. 
If it can't be added to boot2docker (which should auto-detect the Vbox mount) we may be able to create a 'sync mounts' in the pref of Kitematic to get it working? `sync mounts` -> `call above script hack` -> `Mount the Volume within kitematic`

extra-0:
Glad to help.
On the other hand I start thinking that we can achieve even more flexibility.

1. Allow selecting folders regardless of whether they are shared
1. As soon as user requests a container to be started, share and mount required folders.
1. As soon as container stops - remove the share (or just leave it be)

There might be some things to keep in mind. For example data-only containers. We should check from volumes-from and mount volumes from target container as well. We should also keep a registry of folders we have mounted to handle the situations when containers use folders that nested (i.e. container1 uses /Volumes/dev, container2 uses /Volumes/dev/bla) or multiple containers can use the same folder in which case we should only remove share when all of the containers that require it are stopped (or just leave the share as it is. It will be removed after boot2docker restarts)

If you like this approach I can create scripts that will add shares and mounts and remove them accordingly.

alexandrev:
Hi @extra-0. I think your new dynamic approach would be great also!. If you create the scripts I could added when you create a new volumen from a item not previously shared! We are going to create a great feature together! So powerful!

FrenchBen:
@alexandrev @extra-0 if you can work on it together that would be awesome! :tada: 
For 'clarity' it would still be nice to see the shared folder in Virtualbox; overall the script does the heavy lifting that we need, without having to mess with too much user-actions (point and click in Vbox manage)

kotowick:
Glad to see progress! Here to help as well.

alexandrev:
Please, MacOS X users, try it because on Windows this problem was already fixed, so we need to know if it is fixed on Mac. 

Keep me posted of any QA fail or something you see wrong!

extra-0:
@alexandrev there's some sort of glitch when I start kitematic for the first time (with docker machine offline). 
screen shot 2016-02-14 at 20 52 34
As soon as I change settings for any of the folders, it reloads settings for other folders as well.
screen shot 2016-02-14 at 20 53 07
Besides that everything seems rather usable and works as expected. If I start kitematic with docker machine online and mounts are still there from previous launch this doesn't happen.

p.s. I go offline for 2 days and there's so much new. keep it up :)

alexandrev:
Ok man, I'm going to see if I can reproduce it on Windows and update to fix the bug! Thx!!

alexandrev:
Please @extra-0 try it again. I've done a small change and I think the problem you commented is fixed.

extra-0:
@alexandrev ok, now it is showing the folders, but it doesn't mount them. I have to click change on each of the paths in order to force-mount them.

alexandrev:
Mmm I could solve that issue easy.. but I think we should face this another way. We should verify and mount the folders before the start container takes place. Becuase it has no sense to go to the volumes folder to have your volumes mounted. I'm going to do this approach and I get back to  you.

alexandrev:
I've updated with the new approach. Please, @extra-0 take a look!  Thx!

extra-0:
@alexandrev Everything seems to be working now. Flexibility we've achieved is superb!
Thanks!

alexandrev:
Thanks to you because without your script, this could never be posible :)

alexandrev:
@FrenchBen maybe you could take a look at this and give us your comments :)

alexandrev:
Any more QA have been done? @jeffdm , what do you think about the feature and the code? Any comment?

extra-0:
I've been using this branch on a daily basis for past few days and encountered no problems. Would love to see this in production.

alexandrev:
Ok!! Thank you @extra-0. @jeffdm and @FrenchBen does it look good for you?

kotowick:
@alexandrev compiled and works well for me - haven't ran into any issues.

Thanks for doing this!

extra-0:
@FrenchBen Any particular reason for this not to be merged? Anything needs to be reworked?

FrenchBen:
@extra-0 I'd like for the solution to be within boot2docker more than a hack from our side to get it working - Having `sudo` anything as part of the script scares me a bit. 

This only solves the kitematic side of things, when this could benefit all other users (cli users) - If it's done accordingly within the boot2docker setup, then everyone can benefit from it.

extra-0:
@FrenchBen I believe we didn't use any hacks. On the contrary, we exploited functionality that was already there in boot2docker. Isn't Kitematic all about automation? We just automated a process of mounting volumes as user changes settings or starts new containers.

It's extremely flexible and by my experience brings Kitematic users to the next level without binding them to use home folders only. Please, check it out!

alexandrev:
@FrenchBen, I don't see the problem of using sudo inside a VM. It's like using it inside a sandbox. Which is the worst thing could happen? the problem will be using it inside the host machine.

@extra-0 Do we have an issue open to boot2docker to handle this issue? Because I think we didn't open any issue there...

extra-0:
@alexandrev I just browsed through your code again and found no mentions of sudo. Am I wrong?

Also, I suppose boot2docker doesn't have an issue with functionality we are using, but rather an issue with remounts after reboot. We could create a ticket in their repository, however mounts that are used by your version of Kitematic handle this in another way by mounting folders when they are needed.

alexandrev:
We have a sudo when we do the mkdir and mount inside the boot2docker
machine.

Yeah, I'm aware our action is different but if @FrenchBen as core maintener
thinks it is better to have the other approach, we should go for it.

Thx man!

extra-0:
@FrenchBen @alexandrev There were several tickets and even a pull-request at boot2docker to fix this, however it has not yet been accepted and this issue does not get much attention in boot2docker.
https://github.com/boot2docker/boot2docker/pull/1088
https://github.com/boot2docker/boot2docker/pull/1086
https://github.com/boot2docker/boot2docker/issues/1101
The worst that could happen is that boot2docker implements automount and @alexandrev 's mkdir and mount command within boot2docker machine wont' work and can be removed. However all of the other functionality would stay the same.
What do you think?

FrenchBen:
Sudo is targeted to run in a 'sandbox' but say something odd happens and the `sudo` runs on the user's laptop? I wouldn't want to be responsible for that. 

Having the fix in boot2docker would make this work across cli and kitematic. Having the desktop app include more functionality than is available via the normal CLI seems a bit conter-intuitive. If you don't mind pinging the devs in your last issue first, it would be appreciated:
https://github.com/boot2docker/boot2docker/issues/1133

If we don't see anything within the next 2 weeks then we can all vote to merge it and provide more power to Kitematic user :D

Agreed?

alexandrev:
I agree @FrenchBen. But I want to talk about the sudo "problem". The sudo
is only targeted to run inside the sandbox. If something odds happen and
the sudo executes on the real machine. Different alternatives:

a) Windows platform: Nothing happens.
b) MacOS X / Linux platforms: Nothing happens because it is going to
request your user's password to executes any command (in the case your user
belongs to the ones allowed to execute sudo on that machine).

Thanks!!

alexandrev:
Hi, 

I want to review my open #PR and see if there is some need to change anything to merge them or close them if they are not needed. I don't know if this PR makes sense right know after the big change of Docker for Windows / Mac and the change from VirtualBox to Hyper-V and the other option on Mac. @FrenchBen do you need something from my side or do we close this PR?

Thanks!

FrenchBen:
I think this can be closed - I submitted a PR that checks for an existing Docker for Mac/Windows install and allow other folders to be mounted based on that. 
Virtualbox will remain 'locked' to `/Users`

campbx:
Everyone, please consider merging in code from @syardumi #376 

Mac OS version is not able to mount any volumes from Kitematic, this change will help resolve that.

alexandrev:
@FrenchBen I think we have to focus on the newer versions Docker for Windows and Docker for Mac. So, if it is solved there. I'm ok with you closing it. But I think I need to push the transition to Kitematic for the new versions. Because right now it is very difficult to collaborate on things when the master repo seems to be more related to the 'old Virtualbox-based Kitematic'

The main idea here, is that all of us have to work on the same direction to do greater things :)

FrenchBen:
@alexandrev absolutely - I merged a couple of PR related to the changes for folders to be reachable outside of the `Users` directory.

jhgorse:
I am on macOS Sierra and it is not clear how to add new volumes as a result of this feature, particularly when virtualbox is not being used.

376 appears to address the issue of adding new volumes to the dockerfile.

FrenchBen:
@jhgorse the above feature only applies to native users. If you're on Docker for Windows, you simply need to use the settings window to add the volumes that should be shared. See:
https://docs.docker.com/docker-for-windows/#/shared-drives

jhgorse:
@FrenchBen I have what I believe to be the native docker-beta for macOS Sierra, which also has the shared directories (default /Users:/Volumes:/tmp:/private).

These do not appear in the Kitematic->Settings->Volumes settings.
**Also, there is no option that I see to "Add a folder."** <-- This is my main concern, which I believe to be related to #376. Even though #376 has been closed, the issue remains in the beta I have, 0.12.0 from July 27th, and in the screenshots above I do not see the capability to "Add a folder".

FrenchBen:
@jhgorse The 'add a folder' screenshot is from Virtualbox - Kitematic will not show any folder if the image that you're using doesn't have a volume directive. 
Try something like the Kitematic nginx hello world and you'll see the folder option. 

Adding a folder when one isn't declared in the image, has yet to be implemented with the latest changes. This was started in #722 which was closed in favor of this one, which solved only part of the issue.

LiamKarlMitchell:
Still getting error.

image

I deleted the machine in virtual box.
Then ran this in the Docker Quickstart Terminal.
```
docker-machine rm default
docker-machine create --driver virtualbox --virtualbox-share-folder "e:\Docker:/c/project" default
```

If I just patch the validation checks in the UI files would it let me mount volumes from that directory? :)

Note: I am wanting to use VirtualBox on Windows 10 as I can't seem to use VBox with the native windows emulation at the same time and I have many VBox vm's. And the native windows way didn't work well with my computer/network card for some buggy reason it was unbearably laggy. (Broadcom)

PR #1217, opened 2015-11-14
Feature Request: Add exposed ports using the GUI

### LINKED ISSUE #1206: Add exposed ports using the GUI - Feature Request
Today I notice I can not only change the port number of the local side but the protocol too. I can change from a XXXX TCP port to a XXXX UDP port, but I cannot add a new port. 

It doesn't seem more difficult to change the current port mapping than add the port protocol in the current ones. 

Could be posible to add these as a new feature form the app?

alexandrev:
Fixes #1206 . In this PR we allow to add new ports using the Ports tab inside the Settings section. It also allows to remove them using the same form.

We include a button with an icon to remove and add ports as needed and the GUI follows the same approach that the one to add environment variables.

Signed-off-by: Alexandre Vázquez <alexandre.vazquez@gmail.com>


FrenchBen:
I do like it, although the UX is a bit weird.
When I fill-out the ports and I hit 'Save', without clicking '+', I expected the port to be added. It was instead discarded.

When playing with ports, it's not possible to remove the EXPOSED ports from the list - Perhaps this should be checked prior to adding the 'X' (delete) option? Essentially only allowing a user to delete the ports they manually added

Something else I noticed is that when adding an extra port and stopping the container, the originally exposed port-mapping is 'stuck' and will never change, when it should be released.

You can easily replicate by adding a new port to the 'kitematic/hello-world-nginx' and stopping it. Then do a docker inspect on it. 
Compare that to a similar port assignment with the current kitematic version

alexandrev:
Ok, I think you are right with the UX issues you are describing. I'm trying to work it out as soon as possible and I come back with another version :).

FrenchBen:
:clap:

alexandrev:
I already commit the fixed for the first issue but I'm not finding any differences between the "initial exposed port" and the newly added exported ports to discriminate when show the "delete icon". I'm going to keep trying to find a way, but no luck for now.

And I really don't understand the last problem. I add the port, stop the container and I see the "PortBinding" already there, is that the problem that you are talking about? Because I could change the value without anyproblem so I don't feel that the value is stuck.

I want to thank you for your help and support as always :)

FrenchBen:
For the port discrimination you may need to compare 'PortBindings' and 'ExposedPorts' - ExposedPorts should always have what's in the original Dockerfile, while PortBindings will have the 'additional' port I believe.

New error: When I hit 'save' the port list disappears until I leave the tab and goes back to it.

Regarding the last problem, you will need to build 2 containers. 1 where the additional port is added via the command line and another one where you add the port via Kitematic. 
Both output of `docker inspect` should match when the container is running or stopped.
```
$ docker run -itd --name hello1 -P -p 888 kitematic/hello-world-nginx
$ docker run -itd --name hello2 -P kitematic/hello-world-nginx
$ docker inspect hello1
$ docker inspect hello2
```

alexandrev:
Regarding the first thing, the configuration you commented it is not the same I've been facing tonight when I tried to fix that bug. It is true that the first execution I only have the "dockerfile ports" in the ExposedPorts, but when I added a new port the ExposedPorts updates with this definition making impossible for now figure out a way to do this discrimination.

alexandrev:
Thanks about the clue regarding the last problem I'm going to check it out! Thanks a lot!

alexandrev:
I've doing several test as you commented with this result:

If I execute the followinng command:
```docker run -itd --name hello1 -P -p 8888:8888 kitematic/hello-world-nginx```

and then I execute a ```docker inspect hello1``` and I can see as exposed ports both of the ports (the one that exists in the Dockerfile and the one we are specifying in the docker run command:
```
      "ExposedPorts": {
            "80/tcp": {},
            "8888/tcp": {}
        }
```

And then if I execute docker stop hello1, and the docker inspect I get the same output as I get using the Kitematic (with the PR):
```
  "PortBindings": {
            "8888/tcp": [
                {
                    "HostIp": "",
                    "HostPort": "8888"
                }
            ]
        },,

```

So I have two conclusiones from here:

- I cannot reproduce the issue @FrenchBen reported: "Something else I noticed is that when adding an extra port and stopping the container, the originally exposed port-mapping is 'stuck' and will never change, when it should be released." 
- I have to find another way to discriminate the ports declared inside the dockerfile from the ones that you can add. I'm thinking about to recovering the image again and check against it, but I don't know if it too heavy operation, I'm going to give it a try! 

Anothing I cannot reproduce on Windows the problem @FrenchBen said: "New error: When I hit 'save' the port list disappears until I leave the tab and goes back to it." When I hit save, the input text are disabled while the container is restarting and then the input text comes editable again, but for no moment the port list disapear from my watch.

FrenchBen:
@alexandrev I can reproduce the issue consistently. Here's how:
* Create a new container based on the hello-world-nginx
* Go to the settings -> Ports tab.
save-step-1

* Enter a new port number, say 888
save-step-2

* Don't hit save, but click on the '+' to add the port to the list.
save-step-3

* Hit save and notice that the port you added via the '+' will disappear until you go back to a different tab and come back to the port tab
save-step-4

FrenchBen:
Similarly, if you follow the above steps, but don't click the '+' and just hit save, the port will not be saved?
You can check this by clicking 'Save' (the list gets updated as-if it worked), go to the 'General' tab, then come back to the 'Ports' tab and notice that the port you added (without clicking '+' but by clicking 'Save') is no longer part of the list.

alexandrev:
Sorry, now following your instructions could reproduce the issue. I'm going to work on it right now to solve it. My bad

FrenchBen:
no problem. Thanks for all of the work on this!

alexandrev:
Now the initial ports have no option to remove them from the configuration:

image

FrenchBen:
:clap: I'll do some qa on this today. Thanks again for all of the troubleshooting

alexandrev:
Hi @FrenchBen, I think the squash it's done but please take a look at it , because it is the first time that I do something like that. 

On the other side, It's a pleasure to collaborate with you and I'm ready to do all my best to improve Kitematic :). It's the least we can do to support this great app

alexandrev:
I dont have any news about these PR's and they already are facing conflicts because of the changes done in the master. Any way to keep helping you with these developments?

FrenchBen:
Can you update the PR and we'll have another look at it?

alexandrev:
When you talk about updating the branch, are you talking about merging the
changes from the current master to this branch to get an updated version of
the PR?

Thanks in advance

FrenchBen:
That's correct - This way we fix the error of 'This branch has conflicts that must be resolved'

alexandrev:
Another PR updated to the changes done in the master branch. Feedback welcome as always.

FrenchBen:
I'll check it out tomorrow - As with before, if you don't mind squashing the commits, it would be appreciated!

alexandrev:
I've updated the PR with your latest updates from the master branch. I'm also try to squash all the commits in one, but with all the merges and so on, this is the best I could do so I hope the final result its enough to advance this PR.

alexandrev:
No news?

FrenchBen:
Been busy fixing bugs and overall enhancements, which unfortunately take precedence over additional features.

alexandrev:
Ok, I understand, don't worry! Please, let me know if I could help with
something else! (Bugfixes or new feature, whatever thing you need from me)

jeffdm:
Sorry for the delay! The diff seems to be a `console.log` statement. Is the change missing?

alexandrev:
Obviously there is some problem. Let me check and I'll come back to you. Sorry for bothering you

alexandrev:
I lost the changes so I'm creating a new PR with the latest master. It's already created: https://github.com/docker/kitematic/pull/1447

PR #1223, opened 2015-11-17
Issue 1211 - Customizing command to start the container

alexandrev:
First attempt to customize the CMD command when you have a container added to Kitematic.


FrenchBen:
Not sure if this is still work in progress? If it is, please don't create a PR until it's done.

Right now when I go to view a container's command (eg kitematic/hello-world-nginx) I get:
`sh,/start.sh`
What I expected was the same output as the `docker ps`:
`sh /start.sh`

alexandrev:
Hi @FrenchBen , you are right. Now, you are seeing sh,/start.sh because the Cmd parameter inside the container is an Arrays as the CMD reference says and I'm printing the array without transformation. If you prefer the transformation I could do it to see in the GUI without the comma printed and then when you save it again create the array again. 

I'm going to try to make this change today so you can try it out!

Once again, thanks a lot for your time :)

alexandrev:
Hi @FrenchBen,

I commit the fix you were talking about. As you can see in the screenshot, now it displays the correct way:

image

The only thing it is not worked as expected (but I thinks it's better this way) is that if you put only a command like '/bin/bash' and you save it, the current command saved, executed and the one that displays back to you its something like /bin/sh -c /bin/bash.

But I think it's better this way because this is the normal way to specify a command when you start a docker container.

FrenchBen:
Thanks for the quick turn-around, I'll take a look at this today and will report back on behavior

alexandrev:
Any new feedback about this PR? Thanks a milion!

FrenchBen:
It behaves as expected - My concern is passing a command that doesn't exist which would kill the container. I don't know if there's a way to really check if the command exists?

What are you hoping to accomplish with the ability to switch the command from within the GUI?

alexandrev:
I don't think to check the command exist inside the container would be posible. My goal with this PR is to increase the flexibility options to work with containers and try to avoid to use the CLI to do this kind of tasks that is usually when you are working with containers to override the CMD definition or to avoid an error in the Dockerfile itself.

FrenchBen:
@alexandrev I like the idea of modifying a running container and could see the utility (amongst sending people to the cli) but I think the benefit would be much greater if this was an option that the user could provide prior to creating the container. 

There are a few images that do not have a CMD/ENTRYPOINT which then fail in Kitematic. Adding support in the container settings helps, but having it prior to the container creation would be huge!
I'll open an issue to see if people would consider this an improvement

alexandrev:
I agree with you that that will be a huge improvement. Do you want to make it inside this PR or with another PR to that issue?

FrenchBen:
I created the following issue: 
https://github.com/docker/kitematic/issues/1260

Let's discuss it there and we can then enhance your PR :D Thanks for all of your contributions to this, it's really nice to see these coming in

alexandrev:
Perfect!!

jeffdm:
We'd also need to support `ENTRYPOINT`

alexandrev:
Hi @jeffdm , But you are talking about following the same approach you can see in this PR? Like another input field to change the ENTRYPOINT? Or are you talking about any other way to do it? Like the ways we are talking about in the issue: #1260

alexandrev:
I update the PR with your latest updates from the master branch. I'm also try to squash all the commits in one, but with all the merges and so on, this is the best I could do so I hope the final result its enough to advance this PR.

FrenchBen:
A cleanup of the commits would be nice, just to keep our Git history sort-off clean. 
As a heads up, there's a pending PR to update the CMD a bit more:
https://github.com/docker/kitematic/pull/1340

alexandrev:
@FrenchBen I try to clean the branch but I'm not capable to remove the merge commits :(. Regarding to the other PR you talked about (#1340) I don't see any improve about the behavior of this PR. I already include the change @jeffdm  did regarding the Entrypoint. 

I also aware he have the issue opened https://github.com/docker/kitematic/issues/1260 to think about better handling of this feature but as it seems not a lot of "votes" on this I think at least we (when I said we, I said you as maintainer) should go forward with this PR, but that's only IMHO. :)

Thanks a million once again.

FrenchBen:
As long as your own comments are merged, then we're fine- The point of the vote is to make sure we don't include something that not many want and end up having to maintain it (we maintainers).

The balance of adding features vs the need to have the feature is what slows some of these PR :[ - I need to get caught up on the year end stuff and will review this again with jeffdm to make sure we're all on the same page. 

Thanks again for all the work!

alexandrev:
I've been taking advantage that this PR is not very "alive" to do some improvements to it. So, please take a look at it.

I introduce the posiblity to specify the initial command to start the container, you can see the screenshot here:

image

I have to use another library to create the dialog because the electron API we have been used doesn't support it. I try to open an issue to the atom/electron API: https://github.com/atom/electron/issues/4134

This behavior is the explained below:

- When you click on "create" the pop-up is showed.
- If you leave the pop-up content as empty or 'default', the container will be created using the defaults parameters.
- If you specify any command the container is launched with that command.

That's all from my side.

Feedback?

Thx!

FrenchBen:
Closing this until we agree on the flow from: https://github.com/docker/kitematic/issues/1260 We can then start fresh or re-open

alexandrev:
Ok. As you wish :)

PR #1241, opened 2015-11-23
Add package option for devTools window on start

FrenchBen:
Simple command line change to allow the dev tools to be open when the node env is set to development.
Can be started via command line: `npm run start-dev`


jeffdm:
LGTM!

PR #1295, opened 2015-12-13
Fix bug where renaming or editing container settings would cause volume settings to be lost

jeffdm:


mchiang0610:
LGTM on Mac - El Capitan 10.11.2

PR #1301, opened 2015-12-14
Fix bug where container would be run with bash even thought it has a valid entrypoint

jeffdm:


mchiang0610:
LGTM.

src/utils/DockerUtil.js
```diff
@@ -126,7 +126,7 @@ export default {
         return;
       }
 
-      containerData.Cmd = image.Config.Cmd || 'bash';
+      containerData.Cmd = image.Config.Cmd || image.Config.Entrypoint || 'bash';
```
FrenchBen: Unfortunately I don't think this is correct - Doesn't Passing the entrypoint as the cmd to run result in the final cmd be:
`entrypoint + entrypoint`
Previously it was:
`entrypoint + bash`
When no cmd is found, but an entrypoint is, It should just be
`entrypoint`

PR #1321, opened 2015-12-23
Docs

kduraira:


FrenchBen:
When I go to: https://docs.docker.com/mac I see the proper Mac page. Why the need to change the docs?
Additionally it seems that some test fail. For this PR to be considered all tests must pass.

PR #1325, opened 2015-12-27
Feature to change container log font size

aquascaper:
I wrote some code to change the font size for the container logs. I think the buttons can look a lot better and all comments are suggestions are super helpful. 
Signed-off-by: Arjun Patel <arjun453@gmail.com>

### DIFF
--- a/src/components/ContainerHomeLogs.react.js
+++ b/src/components/ContainerHomeLogs.react.js
@@ -35,6 +35,21 @@ module.exports = React.createClass({
     containerActions.active(null);
   },
 
+  getLogFontSize: function(offset) {
+    var fontSize = parseInt($('.logs').css('font-size')) + offset;
+    return (fontSize < 8 ? 8 : fontSize);
+  },
+
+  increaseLogFontSize: function() {
+    var fontSize = this.getLogFontSize(2);
+    $('.logs').css("font-size", fontSize);
+  },
+
+  decreaseLogFontSize: function() {
+    var fontSize = this.getLogFontSize(-2);
+    $('.logs').css("font-size", fontSize);
+  },
+
   render: function () {
     let logs = this.props.container.Logs ?
         this.props.container.Logs.map((l) => <div key={l.substr(0,l.indexOf(' '))} dangerouslySetInnerHTML={{__html: convert.toHtml(escape(l.substr(l.indexOf(' ')+1)).replace(/ /g, '&nbsp;<wbr>'))}}></div>) :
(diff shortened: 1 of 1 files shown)


FrenchBen:
It works well for me - Any chance we could update the `+` and `-` to something a bit more standard like these 
Font icon

aquascaper:
@FrenchBen Are you suggesting that we should remove the buttons on the container log in favor of the drop down menu items and keyboard shortcuts?

FrenchBen:
Yes, I would remove the buttons to keep the interface clean. 
You're correct, it may need to become a separate utility perhaps in `Util.js` to keep the logic of increase/decrease log font size. 

If we pass it as a prop to `logWidget` in the `ContainerHome` file, then it should refresh properly.

aquascaper:
Sounds good! I'll have something for ya sometime next week.

FrenchBen:
@aquascaper Since the log font functionality is going to change a lot from this PR, I'm closing it in favor of a future PR with the above implementation.

aquascaper:
Sounds good! I started traveling recently which made me slow on this PR. I
will open a new one soon.

FrenchBen:
No worries - Since this isn't a critical fix and more of a feature, you can take as long as you need :D

dmitriz:
Just started using Kitematic -- seem like really great tool!

But sadly -- yes -- the font size is very uncomfortably small!!
I have impaired vision and immediately feel eye strain when trying to read it. 

Not only the logs, also Container list, Settings, Volumes etc.
Those are even more important than the logs, because I can copy-paste the latter to view in my editor but can't do the same with settings.

There is huge whitespace wasted, and then the useful content is crammed into tiny 1/10 of available space with tiny fonts. Why?

Letting user scale all fonts together as she likes, would really go far in terms of usability and accessibility! It is just a simple way of making the whole experience less painful and more enjoyable.

I see the objection right now is the presence of those '+/-' in the screen. But these would not add any noise because there is mostly whitespace around.

Please, please, don't ignore this -- it is a huge accessibility issue for people who are older or have impaired vision. They will be thankful for it!

FrenchBen:
@dmitriz I completely agree that this is something that needs to be changed - If you don't mind, create a separate issue with a full feature proposal. There's a lot of work to do and could use the help if this is something you could implement. 
Thanks!

adrian7:
I think would be better to have icons magnifier plus and magnifier minus instead of just plus/min.

aquascaper:
@FrenchBen I understand how to add the hotkeys in `menutemplate.js` and how to pass a prop to `logWidget` in the `ContainerHome' but I am unsure on how to give `ContainerHome` the font size offset value from `Util.js`. Can you give me a pointer?

FrenchBen:
@aquascaper I'm not really sure either, hence this being a non-trivial fix. 
Perhaps you could set something with `localStorage` ?

hellonearthis:
On my 4K monitor and bad eyes makes it a little small.  

Changing font-size in container-home.less details-panel.home .content .mini-logs .widget .logs 
from 10px to 12px works well _but it also changes the search bar font sizing._

PR #1340, opened 2016-01-02
Fix bug where entry point would be repeated if no CMD was specified

### LINKED ISSUE #1339: CMD passed to container is duplicated ENTRYPOINT
It appears a bug was introduced on December 19th which will affect any image which does not define a `CMD` but does define an `ENTRYPOINT`.  In such cases, the `ENTRYPOINT` will be duplicated, often leading to recursive behavior (since the `ENTRYPOINT` does not expect itself to be passed as its own argument).

The exact change which introduced this problem is: [Fix bug where container would be run with bash even with a valid entrypoint](https://github.com/docker/kitematic/commit/ad0d7e003ff4a886c9e8c0abcf274bc837bdc27c)

I created a simple container to illustrate the problem:

Dockerfile:

    FROM alpine:3.2
    COPY start.sh /
    ENTRYPOINT ["/start.sh"]

Contents of start.sh (assuming permissions `chmod 755`):

    #!/bin/sh
    if test "$1" = ""; then
      echo No command was provided.
      exit
    fi
    echo Command provided on command line is: "$*"
    $*

I have built a version of this ([garywiz/entrypoint-bug](https://hub.docker.com/r/garywiz/entrypoint-bug/)) using the above to illustrate the problem.

When run from the docker command line, the following is output, since there is no `CMD` directive, and therefore `start.sh` receives an empty command line, as expected:

    $ docker run --rm garywiz/entrypoint-bug
    No command was provided.
    $

However, if you execute the same image from Kitematic, you see the following output on the console:

    Command provided on command line is: /start.sh
    No command was provided.

Essentially, the `ENTRYPOINT` is being called with itself as an argument, and the second line is output by the recursive invocation.

Upon examination of the December 19th change, Kitematic is building a `run` command that looks like this instead of the correct command which has no argument:

    docker run --rm garywiz/entrypoint-bug /start.sh

There are much more complicated containers which illustrate this behavior, but I created a simple degenerate example to make it clearer.

jeffdm:
This fixes #1339, and does not incorrectly set CMD to ENTRYPOINT if no CMD directive was specified.

### DIFF
--- a/src/utils/DockerUtil.js
+++ b/src/utils/DockerUtil.js
@@ -117,8 +117,13 @@ export default {
       if (!containerData.HostConfig || (containerData.HostConfig && !containerData.HostConfig.PortBindings)) {
         containerData.PublishAllPorts = true;
       }
+      
+      if (image.Config.Cmd) {
+        containerData.Cmd = image.Config.Cmd;
+      } else if (!image.Config.Entrypoint) {
+        containerData.Cmd = 'bash';
+      }
 
-      containerData.Cmd = image.Config.Cmd || image.Config.Entrypoint || 'bash';
       let existing = this.client.getContainer(name);
       existing.kill(() => {
         existing.remove(() => {


jeffdm:
@garywiz thanks for pointing this issue out, and apologies for my incorrect assumption that specifying CMD will override the entry point directive.

Just an update that I've tested this with all three cases:

1) CMD is specified
2) No CMD is specified, but ENTRYPOINT is specified, KM will now correctly run only the entry point command
3) If no CMD or ENTRYPOINT are specified then `bash` is run

jeffdm:
@garywiz @FrenchBen please take a look when you have a moment

FrenchBen:
LGTM :+1:

PR #1362, opened 2016-01-13
make REGHUB2_ENDPOINT configurable

sebadiaz:
I think and i m sure you will not happy. But having the hub url configurable by the properties added to the actual environment variabel could help on some situation where we test more than one hub at the same time

### DIFF
--- a/src/components/Preferences.react.js
+++ b/src/components/Preferences.react.js
@@ -1,5 +1,6 @@
 import React from 'react/addons';
 import metrics from '../utils/MetricsUtil';
+import regHub from '../utils/RegHubUtil';
 import Router from 'react-router';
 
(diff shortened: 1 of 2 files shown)


FrenchBen:
@sebadiaz This has been brought up on multiple occasions and is a bit more tricky than what you have implemented here. 
For the background, see: https://github.com/docker/kitematic/issues/210

sebadiaz:
@FrenchBen Step by Step we will create a better product.
Effectively, actually my backend is just a static json file.

FrenchBen:
@sebadiaz I'm closing this until a proper solution is implemented. Unfortunately this not being complete, we can't merge it.

PR #1363, opened 2016-01-13
Make Recommendations url on kitematic.com configurable

sebadiaz:
I would propose my proper recommendations on kitematic UI.

But I need to change the base url

### DIFF
--- a/src/components/ImageCard.react.js
+++ b/src/components/ImageCard.react.js
@@ -9,13 +9,15 @@ import containerStore from '../stores/ContainerStore';
 import tagStore from '../stores/TagStore';
 import tagActions from '../actions/TagActions';
 import numeral from 'numeral';
+import regHub from '../utils/RegHubUtil';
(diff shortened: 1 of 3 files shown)


FrenchBen:
I like the code, but unfortunately don't see a many use cases in customizing the recommended images. 
The recommended images are official repos and what 99% of the users look for.

sebadiaz:
@FrenchBen I m not agree on the choice of the recommendations. Depending on your interest. Exemple in games: I would have suggestion on konquest-server,chess-engine, chess-rating,counter-strike-server, poker-server, ....  

I know that the server is not ready, But I use a simple static json file. And the url point to docker hub , for resolving the image pulling ans content

FrenchBen:
@sebadiaz Unfortunately this isn't a very common use case - The choice of recommendations could be solved by simply displaying the user's Star repo from their hub account.

PR #1371, opened 2016-01-16
Update year range to 2016

pra85:


Gruntfile.js
```diff
@@ -91,7 +91,7 @@ module.exports = function (grunt) {
             'FileDescription': WINDOWS_APPNAME,
             'InternalName': BASENAME + '.exe',
             'OriginalFilename': BASENAME + '.exe',
-            'LegalCopyright': 'Copyright 2015 Docker Inc. All rights reserved.'
+            'LegalCopyright': 'Copyright 2016 Docker Inc. All rights reserved.'
```
thaJeztah: Should probably be 2015-2016 (or simply "Copyright Docker Inc....")

thaJeztah: http://www.contentious.com/2007/01/07/copyright-notice-is-the-year-really-necessary/

pra85:
@thaJeztah  I have removed years from the copyright notice. But I would also suggest getting advise from somebody more knowledgeable about this subject (The reason why I am suggesting this is because, the site you mentioned above - http://www.contentious.com/2007/01/07/copyright-notice-is-the-year-really-necessary/ - does have a year in their footer region, the thing which they claimed to have removed site-wide )

thaJeztah:
Alright if we really want to be sure, then I suggest making it 2015-2016 in Gruntfile.js, and 2014-2015 in the License.

Reason I commented on the Gruntfile.js, was that you changed the year (and that year was the date the file was originally created)

pra85:
I have updated the pull request with the changes you suggested

FrenchBen:
IANAL but LGTM. Ping @jeffdm

jeffdm:
Thanks! Awesome!

One last thing, can you sign your commit?

```
$ git commit --amend -s --no-edit
$ git push -f
```

pra85:
@jeffdm I have signed my commit and updated the pull request

jeffdm:
LGTM, thanks!

PR #1394, opened 2016-01-20
Fix stop the machine on close on Windows

alexandrev:
Introduce a delay and a two-step close windows procedure to allow to receive and execute the order to stop  the machine when Kitematic closes on Windows.

### DIFF
--- a/src/browser.js
+++ b/src/browser.js
@@ -7,10 +7,12 @@ import os from 'os';
 
 import path from 'path';
 import child_process from 'child_process';
+let Promise = require('bluebird');
 
 process.env.NODE_PATH = path.join(__dirname, 'node_modules');
 process.env.RESOURCES_PATH = path.join(__dirname, '/../resources');
 process.env.PATH = '/usr/local/bin:' + process.env.PATH;
+var exiting = false;
 
 var size = {}, settingsjson = {};
 try {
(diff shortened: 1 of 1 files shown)


alexandrev:
Ok, I found a way to solve the issue https://github.com/docker/kitematic/issues/1369 but I have to admit it's not a very 'pretty' solution but it has the only way I found to get it working.

The problem is this:

- On windows you are not using the before-quit event (I don't know why, but it doesn't work the same way than on Darwin system (MacOS X) ) but the close window event. 

When you receive that event and send it a new event to the app.js to close the machine if the setting is checked, but there is no time to receive the event so the action is never executed. So, the idea here is to do it two times. The first time you send the event and set a Promise to execute again the close method and then it closes.

If you find a better idea or way to get it fixed, please, go ahead!

mchiang0610:
@alexandrev Thanks for the notes. We are working on fixes for VM issues for both Mac and Windows. That being said, it'll take us some time to test and release this.

alexandrev:
No problem. I'm here only to help you. Have a nice day

alexandrev:
it seems that we have to prepare a simple app to prove the issue, so this it is going to take longer than expected. Which is the normal procedure here? We freeze the PR until we have a better solution from atom/electrom API? We go forward with this until we have a better solution to solve the UX for the Windows users? /cc @FrenchBen 

Thx!

alexandrev:
I don't understand we are wasting so much time to merge so easy PR to leave it to have conflics with the master branch...

FrenchBen:
@alexandrev I thought that the last iteration made it impossible to determine the closing of a window and that an issue would be opened with Atom? 
Merging code that is a 'temp' fix isn't a good solution IMHO. Unfortunately there are a lot of issues that need to be looked at first before the qa on the PR can be made accordingly. 

Time is the enemy here, which causes the above conflicts unfortunately :[

alexandrev:
Updating the branch to avoid conflicts.

PR #1430, opened 2016-02-02
Set 'Setup Halted' event instead of 'Setup Failed' for exit code 3

nathanleclaire:
cc @dgageot @jeanlaurent @FrenchBen @jeffdm  FYI

Requires: https://github.com/docker/machine/pull/2979

Signed-off-by: Nathan LeClaire <nathan.leclaire@gmail.com>

### DIFF
--- a/src/utils/Util.js
+++ b/src/utils/Util.js
@@ -13,7 +13,7 @@ module.exports = {
     return new Promise((resolve, reject) => {
       child_process.execFile(args[0], args.slice(1), options, (error, stdout) => {
         if (error) {
-          reject(new Error('Encountered an error: ' + error));
+          reject(error);
         } else {
           resolve(stdout);
         }
--- a/src/utils/SetupUtil.js
+++ b/src/utils/SetupUtil.js
@@ -11,6 +11,11 @@ import machine from './DockerMachineUtil';
 import docker from './DockerUtil';
 import router from '../router';
 
+// Docker Machine exits with 3 to differentiate pre-create check failures (e.g.
+// virtualization isn't enabled) from normal errors during create (exit code
+// 1).
+const precreateCheckExitCode = 3;
+
 let _retryPromise = null;
 let _timers = [];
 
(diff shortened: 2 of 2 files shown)


src/utils/Util.js
```diff
@@ -8,12 +8,23 @@ const remote = electron.remote;
...
 
+// Docker Machine exits with 3 to differentiate pre-create check failures (e.g.
+// virtualization isn't enabled) from normal errors during create (exit code
+// 1).
+const precreateCheckExitCode = 3;
+
 module.exports = {
+  errPrecreateCheckFailure:  'Pre-create check failed: ',
   execFile: function (args, options) {
+    var that = this;
```
jeffdm: Should be unnecessary given that we use fat arrow functions. I.e. the  `=> {`  syntax.

FrenchBen: ES6 - You shouldn't need to do 'that' for 'this' :)

src/utils/Util.js
```diff
@@ -8,12 +8,23 @@ const remote = electron.remote;
...
     return new Promise((resolve, reject) => {
       child_process.execFile(args[0], args.slice(1), options, (error, stdout) => {
         if (error) {
-          reject(new Error('Encountered an error: ' + error));
+          if (error.code === precreateCheckExitCode) {
+            reject(new Error(that.errPrecreateCheckFailure + error));
+          } else {
+            reject(new Error('Encountered an error: ' + error));
```
jeffdm: Simple return `error` here via `reject(error)`. Then check `if (error.code && error.code === 3)` in the `SetupUtil` code.

jeffdm:
Great! Thanks @nathanleclaire . Some small comments on my end

src/utils/Util.js
nathanleclaire: :heart_eyes:

nathanleclaire:
Revised my PR to not be such a ES6 / promise n00b.  PTAL

PR #1441, opened 2016-02-04
Revert "added toolbox overview"

londoncalling:
Reverts docker/kitematic#1437

File needs to be in docker/toolbox (not docker/kitematic)


jeffdm:
Whoops sorry for missing this in the review

PR #1442, opened 2016-02-04
Better socket support

FrenchBen:
Updated some of the socket support along with changes to the event listeners. Unused logic was removed to clean up the code

### DIFF
--- a/src/components/Setup.react.js
+++ b/src/components/Setup.react.js
@@ -3,11 +3,12 @@ import Router from 'react-router';
 import Radial from './Radial.react.js';
 import RetinaImage from 'react-retina-image';
 import Header from './Header.react';
-import Util from '../utils/Util';
+import util from '../utils/Util';
 import metrics from '../utils/MetricsUtil';
 import setupStore from '../stores/SetupStore';
 import setupActions from '../actions/SetupActions';
 import shell from 'shell';
+
 
 var Setup = React.createClass({
   mixins: [Router.Navigation],
@@ -30,6 +31,10 @@ var Setup = React.createClass({
 
   handleErrorRetry: function () {
     setupActions.retry(false);
+  },
+
+  handleUseVbox: function () {
+    setupActions.useVbox();
   },
 
   handleErrorRemoveRetry: function () {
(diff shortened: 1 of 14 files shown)


jeffdm:
Taking a look! Sorry for the delay

src/components/Setup.react.js
```diff
@@ -84,14 +95,18 @@ var Setup = React.createClass({
...
-    if (Util.isLinux()) {
+    if (util.isLinux()) {
       if (!this.state.started) {
         deleteVmAndRetry = (
           <button className="btn btn-action" onClick={this.handleLinuxDockerInstall}>Install Docker</button>
         );
       }
     } else {
-      if (this.state.started) {
+      if (util.isNative()) {
```
jeffdm: Best to put this in an `else if` instead of creating another indent.

src/utils/DockerUtil.js
```diff
@@ -20,20 +23,23 @@ export default {
   activeContainerName: null,
 
   setup (ip, name) {
-    if (!ip || !name) {
+    if (!ip && !name) {
```
jeffdm: Why the `&&`?

src/utils/SetupUtil.js
```diff
@@ -56,34 +66,48 @@ export default {
...
-  async nativeSetup () {
+  async setup () {
     while (true) {
       try {
-        docker.setup('localhost', machine.name());
-        docker.isDockerRunning();
-
-        break;
+        if (util.isNative()) {
+          localStorage.setItem('setting.useNative', true);
```
jeffdm: Why store this setting?

If the user starts Kitematic and the socket is available, shouldn't it just use that by default, always? If the socket isn't available, then maybe it should just always use VBox?

jeffdm:
A couple comments!^

src/utils/SetupUtil.js
FrenchBen: This prevents the 'native' check on start-up since that's the default. 
If your socket is present (but broken) saving this preference allows the bypass of the socket check

src/utils/DockerUtil.js
FrenchBen: I only want to throw an error if both the name and the IP are unavailable. When setting up a socket, I don't need a `name`, I can just use `localhost` (although a bogus name could be passed)
When we pass the VBox setup, the`ip` and the `name` are needed to verify the machine and cert location.

FrenchBen:
I'll change the `if/else` on the display message - good call

PR #1447, opened 2016-02-05
Fixes #1206 . In this PR we allow to add new ports using the Ports ta…

### LINKED ISSUE #1206: Add exposed ports using the GUI - Feature Request

alexandrev:
…b inside the Settings section. It also allows to remove them using the same form.

We include a button with an icon to remove and add ports as needed and the GUI follows the same approach that the one to add environment variables.

Signed-off-by: Alexandre Vázquez alexandre.vazquez@gmail.com


alexandrev:
Recreate the PR because all the code has been lost!

jeffdm:
Thanks!

alexandrev:
I'm Always glad to help :)

FrenchBen:
Sorry to bring this back up after so many exchanges - If I click on the `+` without any ports, I get a warning:
`Warning: bind(): You are binding a component method to the component. React does this for you automatically`
Also can't seem to get new ports added to stick - is it still working for you?

alexandrev:
Hi @FrenchBen,
It worked the last time I try it. But no problem I'm going to do a some tests to the PR so see if something is wrong. 

To summarize your issues:

-  The warning message.
- You add the port but it doesn't stick. In this case, which kind of flow are you following to be reproduced on my side?

I have to catch a flight  today but I expect to have some time to do the testing and I'm going to get back to you!

Thanks a million  :)

alexandrev:
@FrenchBen  The issues you reported should be fixed by now. PTAL and come back to me if anything is not worked as expected. Sorry for the delay :)

src/utils/DockerUtil.js
```diff
@@ -280,6 +296,7 @@ export default {
       data.Mounts = data.Mounts || existingData.Mounts;
 
       var fullData = _.extend(existingData, data);
+      console.log(fullData);
```
FrenchBen: This should probably be removed, but everything else looks good - I'll merge finally!

FrenchBen:
@alexandrev After some QA I found one small issue.
If you have a port mapped and try to add the same port mapped to a different endpoint, the refresh fails and only 1 port gets mapped. 

For example:

| Container | Mac/Win |
| ------------- | ------------ |
| 80 | 8080|
| 80 | 8081 |

alexandrev:
No problem, I'm going to try to fix it later today. Thanks for the
insights!!

alexandrev:
@FrenchBen  I couldn't work on that, because I cannot build the latest version on Windows. I've opened an issue related to it. I'm going to try to avoid the problem to be able to fix this bug

alexandrev:
Hi @FrenchBen  I've created a new PR related to this issue, you can check it: https://github.com/docker/kitematic/pull/1724

PR #1448, opened 2016-02-05
Use which to find docker-machine

StefanScherer:
I started to build a [Chocolatey package for Kitematic](https://github.com/stefanScherer/choco-kitematic). But installing VirtualBox, docker-machine and then Kitematic through `choco install` doesn't work yet.

The problem is that Kitematic searches the `docker-machine` binary only with the environment variable `DOCKER_TOOLBOX_INSTALL_PATH`. After installing `docker-machine` with `choco install docker-machine` it is in the user's PATH.

So I enhanced the check to find `docker-machine` in the PATH if no such environment variable is set. The environment variable still has priority over the PATH to keep the same behaviour using the Docker Toolbox.

For Linux the `docker-machine` binary also is searched in PATH so it could be put to another place (perhaps someone compiles it from Go sources).

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,8 @@
     "request-progress": "^0.3.1",
     "rimraf": "^2.3.2",
     "underscore": "^1.8.3",
-    "validator": "^4.1.0"
+    "validator": "^4.1.0",
+    "which": "^1.2.4"
   },
   "devDependencies": {
     "babel": "^5.8.23",
--- a/src/utils/DockerMachineUtil.js
+++ b/src/utils/DockerMachineUtil.js
@@ -24,7 +24,12 @@ var DockerMachine = {
     return 'default';
   },
   installed: function () {
-    return fs.existsSync(this.command());
+    try {
+      fs.accessSync(this.command(), fs.X_OK);
+      return true;
+    } catch (ex) {
+      return false;
+    }
   },
   version: function () {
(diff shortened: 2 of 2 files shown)


FrenchBen:
Although I'm a big fan of Chocolatey, we use the Docker Toolbox to do the install on Windows and set the above variables. 
Similarly on the Mac, we use the Docker Toolbox to do the install instead of pushing for a `brew` solution. 

With that said, I do like the improvements to the code in order to locate the proper binary - I'll QA the changes and see if there is a 'breaking' condition. 

As another side note, while you're in there, can you also update the call to the `existSync`? See:
https://nodejs.org/api/fs.html#fs_fs_existssync_path

xslim:
I vote to include this PR. On Mac, I install docker via Homebrew. And it installs `docker-machine` in the Homebrew path...

FrenchBen:
@jeffdm PTAL

StefanScherer:
Any updates? In the meantime someone else has started a Chocolatey package https://chocolatey.org/packages/docker-kitematic and I showed him this PR.

FrenchBen:
@StefanScherer I'm gona go ahead and merge this as it seems to be working well  - Sorry for the delay

StefanScherer:
Thanks @FrenchBen - no problem ;-)

Issue #1041, opened 2015-09-15
Search log files in ui


aaronrobertson:
It would be convenient if you could search the log file output in the ui.

mchiang0610:
@aaronrobertson Thanks, adding this as a feature request

mikaoelitiana:
I am interested to work on this, @aaronrobertson what would be your suggestion to implement it?

aaronrobertson:
Maybe something that looks like this?
![quickmockup](http://i.imgur.com/Qk1aSSH.jpg)

ClemMakesApps:
@mikaoelitiana I might be able to look into this. Have you started work on this?

mikaoelitiana:
I have you yet. You can go ahead

Le mer. 25 mai 2016 19:03, Clement Ho <notifications@github.com> a écrit :

> @mikaoelitiana <https://github.com/mikaoelitiana> I might be able to look
> into this. Have you started work on this?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/docker/kitematic/issues/1041#issuecomment-221620544>
>
-- 

Mika Andrianarijaona
+261340722394
http://mikaoelitiana.name

FrenchBen:
@mikaoelitiana @ClemMakesApps there's a pending PR with some bugs:
https://github.com/docker/kitematic/pull/1522

PR #1522, opened 2016-03-07
#1041 Add feature to find text in log

ysm001:
This PR adds a new feature allowing users to find text in log.

kitematic-search

- Open search panel: Ctrl (or Cmd) + F
- Find next: Enter
- Find previous: Shift + Enter
- Close search panel: Esc

### DIFF
--- a/src/actions/LogActions.js
+++ b/src/actions/LogActions.js
@@ -1,0 +1,17 @@
+import alt from '../alt';
+
+class LogActions {
+  search(searchText) {
+    this.dispatch({searchText});
+  }
+
+  toggleSearchField(searchFieldVisible) {
+    this.dispatch({searchFieldVisible});
+  }
+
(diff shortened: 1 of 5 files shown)


FrenchBen:
I have a container that's just a redis DB that was shutdown for a while and thus didn't have any logs - This caused the following error: 
`ContainerHomeLogs.react.js:55 Uncaught TypeError: Cannot read property 'length' of undefined` 

The above error prevented me from doing anything but close the app. 
Could you see why this is happening?

FrenchBen:
@ysm001 ping

FrenchBen:
LGTM

ClemMakesApps:
The search is currently case sensitive. Shouldn't it be other way around?

FrenchBen:
Closing in favor of #1728

shinriyo:
I want this feature

PR #1526, opened 2016-03-08
Add url scheme for building images

AnkyrinRepeat:
Mac users can use the url scheme 'kitematic://<repo owner>/<repo name>' to open Kitematic and immediately start building the desired repo. 

Signed-off-by: Colin Wiley <colinwwiley@gmail.com>

### DIFF
--- a/src/utils/DockerUtil.js
+++ b/src/utils/DockerUtil.js
@@ -78,7 +78,6 @@ export default {
     this.placeholders = JSON.parse(localStorage.getItem('placeholders')) || {};
     this.fetchAllContainers();
     this.listen();
-    repositoryActions.fetchPending();
 
     // Resume pulling containers that were previously being pulled
     _.each(_.values(this.placeholders), container => {
@@ -207,6 +206,7 @@ export default {
           return;
         }
         containerServerActions.allUpdated({containers: _.indexBy(containers.concat(_.values(this.placeholders)), 'Name')});
+        repositoryActions.fetchPending();
       });
     });
   },
--- a/src/utils/DockerUtil.js
+++ b/src/utils/DockerUtil.js
@@ -8,6 +8,7 @@ import util from './Util';
 import hubUtil from './HubUtil';
 import metrics from '../utils/MetricsUtil';
 import containerServerActions from '../actions/ContainerServerActions';
+import repositoryActions from '../actions/RepositoryActions';
 import rimraf from 'rimraf';
 import stream from 'stream';
 import JSONStream from 'JSONStream';
@@ -77,6 +78,7 @@ export default {
     this.placeholders = JSON.parse(localStorage.getItem('placeholders')) || {};
     this.fetchAllContainers();
     this.listen();
+    repositoryActions.fetchPending();
 
     // Resume pulling containers that were previously being pulled
     _.each(_.values(this.placeholders), container => {
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -57,8 +57,8 @@ module.exports = function (grunt) {
           asar: true,
           'app-version': packagejson.version,
           protocols: [{
-            name: 'Kitematic',
-            schemes: ['kitematic']
+            name: 'kitematic',
+            schemes: ['Kitematic']
           }]
         }
       },
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -57,8 +57,8 @@ module.exports = function (grunt) {
           asar: true,
           'app-version': packagejson.version,
           protocols: [{
-            name: 'Foobar',
-            schemes: ['foobar']
+            name: 'Kitematic',
+            schemes: ['kitematic']
           }]
         }
       },
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -55,7 +55,11 @@ module.exports = function (grunt) {
           platform: 'darwin',
           arch: 'x64',
           asar: true,
-          'app-version': packagejson.version
+          'app-version': packagejson.version,
+          protocols: [{
+            name: 'Foobar',
+            schemes: ['foobar']
+          }]
         }
       },
       linux: {
--- a/src/actions/RepositoryActions.js
+++ b/src/actions/RepositoryActions.js
@@ -20,7 +20,7 @@ class RepositoryActions {
 
   fetch (name) {
     this.dispatch({});
-    regHubUtil.fetch(name)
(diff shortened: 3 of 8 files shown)


mchiang0610:
@AnkyrinRepeat Thanks for submitting this. We have code that was previously tested and passed security team checks. Can we see if we can re-use that code with your modifications? It will be better on our end. 

Details in this PR: https://github.com/docker/kitematic/pull/383 

We can still use the kitematic:/// url scheme 

Thanks,
Michael Chiang 
mchiang@docker.com

AnkyrinRepeat:
Sure, I refactored it to use the 383 code. The URL scheme is now kitematic://repository/run/<user>/<repo> to bring up the 383 confirmation page to begin running a container.

mchiang0610:
Thanks! I will manually test this. @FrenchBen, @jeffdm or @elesant, mind helping to review the code as well?

jeffdm:
Awesome! Taking a look

src/utils/DockerUtil.js
```diff
@@ -212,6 +213,7 @@ export default {
   run (name, repository, tag) {
     tag = tag || 'latest';
     let imageName = repository + ':' + tag;
+    name = ContainerStore.generateName(name);
```
jeffdm: This happens in URLUtil so I think we can remove it here.

src/utils/DockerUtil.js
```diff
@@ -7,6 +7,7 @@ import child_process from 'child_process';
 import util from './Util';
 import hubUtil from './HubUtil';
 import metrics from '../utils/MetricsUtil';
+import ContainerStore from '../stores/ContainerStore';
```
jeffdm: Similar to below - can most likely remove this

src/components/NewContainerPull.react.js
```diff
@@ -0,0 +1,43 @@
...
+import Router from 'react-router';
+import shell from 'shell';
+import containerActions from '../actions/ContainerActions';
+
+module.exports = React.createClass({
+  mixins: [Router.Navigation],
+  handleOpenClick: function () {
+    var repo = this.props.pending.repo;
+    if (repo.indexOf('/') === -1) {
+      shell.openExternal(`https://registry.hub.docker.com/_/${this.props.pending.repository}`);
```
jeffdm: Replace `https://registry.hub.docker.com` with just `https://hub.docker.com`

src/components/NewContainerPull.react.js
```diff
@@ -0,0 +1,43 @@
...
+    } else {
+      shell.openExternal(`https://registry.hub.docker.com/u/${this.props.pending.repository}`);
```
jeffdm: same as above

jeffdm:
Took a first look - so far so good.

src/utils/DockerUtil.js
AnkyrinRepeat: I put it in DockerUtil.run because when a user opens Kitematic using a URL scheme it will create the name before Kitematic has finished loading preexisting containers so if a user had a hello-world and hello-world-2 container it will overwrite the hello-world container rather than name it hello-world-3.  Would it make sense to only generate the name there or do you have another suggestion on how to deal with that?

src/components/NewContainerPull.react.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react/addons';
```
jeffdm: I think we may be missing the CSS for this component from the older PR

src/utils/DockerUtil.js
jeffdm: Ah, good catch! Can we:

- Remove the call to `generateName` in the `ImageCard.react` component?
- Remove the `name` parameter from the `run` action and util function here?

jeffdm: Disregard the previous comment, it would be best to:

- Call `ContainerStore.generateName` in the `NewContainerPull` component instead of here.

jeffdm:
Took another look with some comments - sorry for the delay.

AnkyrinRepeat:
Thanks for the additional comments, I moved the name generation to NewContainerPull. It looks like all the CSS from PR 383 is still in the master branch in styles/new-container.less.

src/components/NewContainerPull.react.js
```diff
@@ -0,0 +1,45 @@
...
+    var repo = this.props.pending.repo;
+    if (repo.indexOf('/') === -1) {
+      shell.openExternal(`https://hub.docker.com/_/${this.props.pending.repository}`);
+    } else {
+      shell.openExternal(`https://hub.docker.com/u/${this.props.pending.repository}`);
+    }
+  },
+  handleCancelClick: function () {
+    containerActions.clearPending();
+    this.context.router.transitionTo('new');
```
jeffdm: I think it's better to `transitionTo('search')`

src/utils/URLUtil.js
```diff
@@ -0,0 +1,51 @@
...
+
+    // Get the type of object we're operating on, e.g. 'repository'
+    var type = parser.host;
+
+    if (this.TYPE_WHITELIST.indexOf(type) === -1) {
+      return false;
+    }
+
+    // Separate the path into [run', 'redis']
+    var tokens = parser.path.replace('/', '').split('/');
```
jeffdm: Since Kitematic only supports Docker Hub repos, I think it's best that we limit to Docker Hub repos (for now), using a check like this:

```
if (tokens.length > 3) {
  return false;
}
```
to avoid pulling from other repos. For example

`['run', 'kitematic', 'hello-world-nginx'] // kitematic/hello-world-nginx` is ok but
`['run', 'other-registry-host.com', 'namespace', 'reponame'] // other-registry-host.com/namespace/reponame` should not be accepted for now

src/components/NewContainerPull.react.js
```diff
@@ -0,0 +1,45 @@
...
+  },
+  handleConfirmClick: function () {
+    containerActions.clearPending();
+    var name = containerStore.generateName(this.props.pending.repo)
+    
+    containerActions.run(name, this.props.pending.repo, this.props.pending.tag);
```
jeffdm: I think here we need to transition to view our new container being pulled! Add a line like:

`this.transitionTo('containerHome', {name});`

jeffdm:
Sorry again for the delay! Did some end-to-end testing - very cool! A few comments on some edge cases, but overall looking great.

PR #1529, opened 2016-03-08
Changes to the Gruntfile to allow to create a release version on Windows

alexandrev:
Changes to the Gruntfile to allow to create a release version on Windows

P.S. Avoid the execution of shell:sign and shell:zip because on Windows there is no way to execute the sign process to Mac Applications.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -26,7 +26,7 @@ module.exports = function (grunt) {
   var OSX_OUT = './dist';
   var OSX_OUT_X64 = OSX_OUT + '/' + OSX_APPNAME + '-darwin-x64';
   var OSX_FILENAME = OSX_OUT_X64 + '/' + OSX_APPNAME + '.app';
-
+  var IS_WINDOWS = process.platform === 'win32';
   grunt.initConfig({
     IDENTITY: 'Developer ID Application: Docker Inc',
     OSX_FILENAME: OSX_FILENAME,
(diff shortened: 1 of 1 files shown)


Gruntfile.js
```diff
@@ -256,7 +256,11 @@ module.exports = function (grunt) {
   });
 
   grunt.registerTask('default', ['newer:babel', 'less', 'newer:copy:dev', 'shell:electron', 'watchChokidar']);
-  grunt.registerTask('release', ['clean:release', 'babel', 'less', 'copy:dev', 'electron', 'copy:osx', 'shell:sign', 'shell:zip', 'copy:windows', 'rcedit:exes', 'compress']);
+  if (!IS_WINDOWS) {
+    grunt.registerTask('release', ['clean:release', 'babel', 'less', 'copy:dev', 'electron', 'copy:osx', 'shell:sign', 'shell:zip', 'copy:windows', 'rcedit:exes', 'compress']);
+  }else {
+    grunt.registerTask('release', ['clean:release', 'babel', 'less', 'copy:dev', 'electron', 'copy:osx', 'copy:windows', 'rcedit:exes', 'compress']);
```
FrenchBen: is `copy:osx` actually needed here?

alexandrev: No, it's not. My first idea was to avoid the execution of the steps that was not possible to execute on Windows. But I think you're right. So, I'm doing a change to remove copy:osx and to change from electron to electron:windows. You should see now the latest version of the commit.

FrenchBen:
LGTM @jeffdm PTAL

Issue #860, opened 2015-07-24
[Enhancement] Windows: add menu bar


frehner:
Forgive me if I missed this, but I searched this repo and didn't find anything.

On OSX, the menu bar provides access to things such as "About Kitematic" (which looks great with that new PR, btw), "Preferences", etc. However, the Windows version doesn't appear to have any equivalent... Unless I'm missing something?

FrenchBen:
Unfortunately I don't use the Windows GUI, but the menu Preferences in OSX, can also be found within the Kitematic GUI - If you look at the bottom left of the Kitematic window, you can see a small 'gear' icon, which will show you the preferences window.

frehner:
Good to know, thanks.  So that's a workaround for preferences... everything else in the OSX menu bar is still inaccessible though (or very well hidden).

FrenchBen:
Probably won't work - but did you try to just press 'Alt' while Kitematic is open? In windows that usually forces any/all menu to appear

frehner:
Yeah, I think I tried that and nothing happened --  but I'm away from my work/windows environment for the weekend to be 100% sure.

frehner:
Just to clarify, the ALT button doesn't do anything.  Also, another reason why this is a slight annoyance for Windows users: the only way to tell which version Kitematic is currently on is by going to the "Uninstall or Change a program" window and looking at the version there. 

This was tested on both 0.7.6 and 0.8.0-rc4.

pmario:
IMO this is a bug and not an enhancement. There is code, that creates a menu. ... I just doesn't work on windows. see: https://github.com/docker/kitematic/blob/master/src/app.js#L54

Bigguy34:
Ok I started to fix this, check out this [PR](https://github.com/docker/kitematic/pull/1559)

PR #1541, opened 2016-03-14
Add windows hyperv support. 

pmario:
It's based on: floradu88's PR. ... but I changed quite a lot. See info below. 
also see: https://github.com/docker/kitematic/issues/1463#issuecomment-195819642

## Hard coded requirements

- User needs to be part of the `Hyper-V-Administrators` group.
   - User needs elevated rights only once: When VM is created. 
   - see: [comment above](https://github.com/docker/kitematic/issues/1463#issuecomment-195819642) point 3
- The virtual switch needs to be named `KitematicVirtualSwitch`

### Setup Kitematic VSwitch

- Open the `Hyper-V Manager` Desktop App
- If you need Internet Access, you'll probably need to use `External Network`!
- see [msdn Hyper-V on Windows](https://msdn.microsoft.com/en-us/virtualization/hyperv_on_windows/quick_start/walkthrough_virtual_switch) for more info.

  - 03 kitematicvirtualswitch_

### Add user to Hyper-V-Administrators group

- Open your file `explorer`

- right click `This PC`: Manage (You have to be Admin to do this !) 

  - 01 right click this pc manage_

- System Tools: Local Users and Groups: Groups

  - 02 manage groups hyper-v-admins_


pmario:
Not tested, if VBox still works :)

pmario:
this command: ` $(Get-VMSwitch | where {$_.SwitchType -eq "external"}).name` will give us an existing switch, with the right parameters. ... So hard coded names will be obsolete.

FrenchBen:
Doesn't `Get-VMSwitch` require admin? How does this work for non-admin users?

pmario:
No. As I wrote above it only needs `Hyper-V-Administrators` rights. So the user needs to be in this group. If you want to use hyper-v you need to be in this group anyway, since otherwise every command needs a boring elevated popup. .... 

With the setting described above, I could reduce the elevated dialog. It's only needed, because `docker-machine.exe` has a hardcoded dependency. Which IMO is a bug! see my comment at: https://github.com/docker/kitematic/issues/1463#issuecomment-195819642 point 3

FrenchBen:
Thanks for the info - I had previously seen some of the elevated admin rights issues and had to ask. It's nice to see this as not being a problem with the user being in the proper group.

pflannery:
Just some feedback when trying to run this on Windows 10 64bit.

- I get an error first time about 'KitematicVirtualSwitch' not existing.
- once I create the correct switch then it creates a hyper-v vm called 'default' (be nice to have a better name like kitematic or similar if possible)
- after powershell has finished the 'default' vm is booted to command prompt but I get this error screen in kitematic (note: my user is in the hyper-v administrators group) 

image

I tried the delete and retry button in kinematic but same issue.
Then tried `docker-machine restart default` which restarts the vm successfully, but kitematic still complains with above error.

boot2docker has a message  `Device "eth1" does not exist`

image

FrenchBen:
Sounds like something changed in your setup since the VM was created? A missing device could indicate issues with communication - Check your Virtualbox settings and what device it uses to communicate with the VM

pflannery:
I have an existing Fedora 23 Vm on hyper-v that can connect to the internet and it works great. 
But when I try to setup with kinematic it automatically creates an extra switch which will never be able to get on to the internet because you can only share the internet with one switch

By changing the hard coded 'KitematicVirtualSwitch' switch name to use my existing switch name seems to of fixed this issue. 

I now see this:

image

awesome to have this working with hyper-v!!

pflannery:
@FrenchBen thanks

pmario:
> By changing the hard coded 'KitematicVirtualSwitch' switch name to use my existing switch name seems to of fixed this issue. 

@pflannery thx for testing. ... 

IMO a new test needs to be made:

 - search for an external switch with `$(Get-VMSwitch | where {$_.SwitchType -eq "external"}).name` ... and just take it.
 - If no switch fits, point the user to a help page. 

What do you think?

pmario:
>once I create the correct switch then it creates a hyper-v vm called 'default' (be nice to have a better name like kitematic or similar if possible)

@FrenchBen .. I just used the same name, that's already used for VBox. But I also thought, it should be a different name eg: `kitematic-default`, `kitematic-vm`  or just `kitematic` as suggested

It would also create better CLI commands eg: `docker-machine start kitematic-vm` or `docker-machine start kitematic`

pmario:
>awesome to have this working with hyper-v!!

Just to be sure: There is __no__ directory sharing atm, since hyper-v doesn't provide this function.

pmario:
@pflannery I did remove the hard coded vSwitch name. We use the first one we find now. .. Can you test pls? ... It took me quite some time to be able to handle spaces in the name :)

What's still missing is, some more config pages. .... @FrenchBen Does it make sense to go on developing, or is it just wasted time. Is there any chance, that something from this branch gets merged?

pflannery:
@pmario thanks. I just tried this and now get this

image

My internet switch is an internal switch.

pmario:
thx. Yea. we need a configuration page. Guessing will be always wrong :/

FrenchBen:
@pmario Since this is a pretty big change in code, could you make sure to sign your next commit?
See our Contributing guidelines:
https://github.com/docker/docker/blob/master/CONTRIBUTING.md#sign-your-work

pmario:
@FrenchBen  ... I will :)

@pflannery & @FrenchBen ... Since using the first "external" switch wasn't a great success :/ There are 2 possibilities left: 

 - use the first one we find no matter if it works. ... + some info howto fix it somewhere
 - read the names of the existing switches
   - present them to the user, prior to do anything and let them decide

The first solution is the easiest to implement, with a very high probability, that it's wrong. ... We could test the switch and add a message eg: "We are about to use your vSwitch named: 'thats the first one'. Is this right?"  ... 2 buttons. [Proceed]  [Show me moar help]

The second solution is kind of impossible at the moment, (in windows) because we can't stop at this time of the process, because the UI has absolutely no interface except the "loading spinner". ... There is no menu bar to reach the settings page. see: #860

pflannery:
> use the first one we find no matter if it works. ... + some info howto fix it somewhere

would work for me though showing a list is probably ideal.

floradu88:
@pmario LGTM

floradu88:
It looks like there might be an issue with this code:

  hasAdminRights: function() {
    return util.execFile([this.command(), 'Get-VMHostSupportedVersion']).then(stdout => {
      console.log('stdout: ', stdout);
      // To execute the above command you'll need Admin or Hyper-V-Admin rights.
      if ( stdout.toUpperCase().indexOf('TRUE') !== -1) {
        return Promise.resolve(true);
      }
    }).catch(() => {
      return Promise.resolve(false);
    });
  },

I found a PC with Windows 10, Powershell 5.0 and Hyper-V version 6.2.

I am suggesting this change below: 

  hasAdminRights: function() {
    return util.execFile([this.command(), 'Get-VMHost']).then(stdout => {
      console.log('stdout: ', stdout);
      // To execute the above command you'll need Admin or Hyper-V-Admin rights.
      if ( stdout.toLowerCase().indexOf('InvalidOperation') === -1) {
        return Promise.resolve(true);
      }
    }).catch(() => {
      return Promise.resolve(false);
    });
  },

floradu88:
@pmario Also looks like if you aren't running as admin and you are in the group still does not work. I will create a fix in a few hours for this. 
Could you please take a look at the previous comment?

pmario:
>@pmario Also looks like if you aren't running as admin and you are in the group still does not work.

You are right. I'm not running as admin. See the OP. I'm part of HyperVAdministrators group. 

>I will create a fix in a few hours for this.
Could you please take a look at the previous comment?

IMO there is no real fix atm. docker-machine has a hardcoded test for admin so atm we have to activate elevated rights to satisfy docker. 

and hyper-v seems to have a problem too: https://github.com/docker/machine/issues/2989

--------------

I'll have a look at your code above

pmario:
>```
if ( stdout.toLowerCase().indexOf('InvalidOperation') === -1) {
```

imo This code won't work, since there is `toLowerCase` and the test string has upper case chars. 

-----------

I don't see the difference between your suggested code and mine. Both commands need Admin or HyperVAdmin rights to execute. ... So if we get a response, the users have the right permissions. 

As I wrote docker-machine has a hardcoded test. So we need to start with elevated rights. That's why we need this super ugly construction here: https://github.com/pmario/kitematic/blob/add-hyperv/src/utils/DockerMachineUtil.js#L79

**Comment by pflannery** [comment]
@pmario Just pulled the latest from here and powershell keeps repeating the following

```shell
default) Calling .GetSSHKeyPath
(default) Calling .GetSSHUsername
Using SSH client type: external
Using SSH private key: C:\Users\DevUser\.docker\machine\machines\default\id_rsa (-rw-rw-rw-)
<nil>
Error getting ssh command 'exit 0' : Permissions 0666 for 'C:\Users\DevUser\.docker\machine\machines\default\id_rsa' are too open.
```

pmario:
It seems they have a new version of boot2docker. I'm testing right now.

pflannery:
> It seems they have a new version of boot2docker. I'm testing right now.
yeah I think your right.

floradu88:
@pmario I don't have this command available "Get-VMHostSupportedVersion" out of the box.
@pmario Yes that code was done without testing it. I only researched available commands that you can use.

pmario:
@floradu88 ok. I see. 

We could use a similar test as docker machine. .. but they use the wrong one atm :/ see: https://github.com/docker/machine/blob/b6be1b79bcb2bf7ef6f74031ed5da3e72b882b4b/drivers/hyperv/powershell.go#L72

pmario:
@pflannery ... for me everything is broken atm. It downloads b2d-rc3 ... at least it thinks so. It checks versions and starts downloading again. ...

pflannery:
@FrenchBen Seems like kitematic is trying to grab a b2d release candidate which is unstable. Is there a way to prevent this from happening ?

pmario:
>@FrenchBen Seems like kitematic is trying to grab a b2d release candidate which is unstable. Is there a way to prevent this from happening ?

It actually is docker-machine. If you try this command `docker-machine.exe -D create --driver hyperv --hyperv-memory 2048 --hyperv-virtual-switch 'switchIntern' default` in Admin Powershell, you'll get the same behaviour.

pflannery:
@pmario ok thanks

FrenchBen:
@pmario @pflannery what version of docker-machine are you running?
`docker-machine version`

pmario:
@FrenchBen 
docker-machine.exe version 0.7.0-rc2, build 60dbecb atm.

pmario:
docker-machine rc3 works

pflannery:
> docker-machine.exe version 0.7.0-rc2, build 60dbecb atm.

@FrenchBen same version for me too

floradu88:
Could we check and merge this in?

pmario:
Hi folks, 
I'm closing this PR since it is (kind of) superseded by DockerForWindows and Kitematic auto detects hyper-v now. 
have fun!

floradu88:
@pmario Hello, maybe next time is worth mentioning that before you spend time developing and test it. Could you please also include the link to the two things you mentioned above?

pmario:
@floradu88 
>Hello, maybe next time is worth mentioning that before you spend time developing and test it. 

Please tell me, how I should have known this? ... 

I still think the Kitematic / docker-machine "multi Platform / single software approach" has an advantage over the "locked in" native versions.

------------

Here's the timeline as I see it. That's my point of view and may be totally wrong, since I don't have any knowledge about the real timeline and the people involved. 

We probably started our approach to include hyper-v into Kitematic at about the same time frame as Docker created D4W. ... D4W concepts may be several months earlier. 

January
The ("unikernal guys")[https://blog.docker.com/2016/01/unikernel/] joined docker in January 

February
Docker (moved closer to alpine in February.)[https://www.brianchristner.io/docker-is-moving-to-alpine-linux/] .. which (btw) caused some confusion in the "ubuntu camp"

(fast forward May 2016: those groups seem to be responsible (for a big part of the implementation)[https://blog.docker.com/2016/05/docker-unikernels-open-source/])

March
This PR started 14th March. 

The first public post in the D4W discussion forum was created on (17th March.)[https://forums.docker.com/t/docker-for-windows-issue-feedback/7591]

Docker for Mac and Windows beta was (announced at the end of March.)[https://blog.docker.com/2016/03/docker-for-mac-windows-beta/] ... I did request an invitation the same day.

April
April 3rd I got my link to test the beta. D4W was far away, from what it is now. ... 

May 9th 
This PR got the "postpone" label, which for me is very close to "no". In the sense of: "No" is temporary and "Yes" is forever ;)

May 10th
(Kitematic 0.10.3)[https://github.com/docker/kitematic/releases] is the first version that supports Docker for Widnows natively. 

You may find a "Docker for Windows" link here: https://forums.docker.com/t/docker-for-windows-beta-invites/9234/4 ...  I don't know if it still works as intended.

floradu88:
@pmario I am not happy about this going to be wasted time in developing something that won't be used. They could have rejected this before we starting to put time and effort into it. A lot of guys like @FrenchBen  and you were involved and everybody looked happy about this being done. But maybe I was wrong to think that.

FrenchBen:
@floradu88 I completely understand the feeling and don't think the effort is completely wasted. 
As @pmario mentioned the timing was less than ideal, as a lot of things happened almost concurrently. 

One REALLY nice use-case for this PR would be to create a Windows server 2016 instance within Hyper-V and allow kitematic to communicate with it. 
Allowing everything/everyone to slap any OS within HyperV and expect Kitematic to just 'work' is a bit more of a stretch, as each my have their own set of features/settings. 

I'd be happy to pull this branch in under the main repo to make sure that the code/time spent remains relevant and doesn't go away (in case repo changes are made by either of you)

You both also have my email, so feel free to ping me offline if needed  :)

floradu88:
@FrenchBen I will be more than happy if this gets merged. @pmario what do you think about this?

pmario:
@floradu88 As I did understand the comment from FrenchBen, he would pull the branch into the main repo, so it doesn't get lost. There wasn't the intention to merge it? @FrenchBen Or am I wrong here?

I don't have the intention to pull a windows server from the web. With my internet connection it takes hours not minutes to pull it. ... I think the ubuntu base images are way to big to be convenient. I use busybox, debian or now alpine. They still need minutes to be initialized the first time. In my opinion pulling gigabytes of images is just wasted time and super boring.

floradu88:
@pmario Yes, pull it under the main repository.

FrenchBen:
Pulling it under the main repo, not merging :)

@pmario if you feel incline, here are some of the TP5 setup steps:
https://forums.docker.com/t/windows-server-2016-tp5-docker-server-remote-management/10315/6

floradu88:
@FrenchBen I am sorry to say this, but the latest docker version is not working out of the box:
"Error: getaddrinfo ENOTFOUND docker.local docker.local:2375" with hyper-v, it fails and suggests to use virtualbox. Any idea why?

FrenchBen:
@floradu88 The windows check for the Docker for Windows isn't the best - If you have suggestions on how I could make it better, I'm open to suggestions.
The issue is detecting whether the user forgot to enable Docker for Windows or it simply doesn't exist. 
Right now it's setup to try and assumes the user forget to start it. 

Thoughts?

floradu88:
I will start from master and debug it to see why this is failing. Will create a pull request with fixes, if I manage to fix it.

PR #1544, opened 2016-03-14
fix for displaying no description on repo containers that had short descriptions

Bigguy34:
Signed-off-by: Thomas Florin <thoflo002@gmail.com>


jeffdm:
LGTM

@FrenchBen PTAL

PR #1546, opened 2016-03-14
Grammar fix, s/seem/seems/

saltybeagle:
The subject of the sentence `an error` is singular, therefore `seems` should be used instead of `seem`.


FrenchBen:
Thanks for catching this and correcting the typo via a PR :+1:

Issue #231, opened 2015-03-12
Don't install VirtualBox by default — use existing installation


develar:
I cannot use VirtualBox because I don't have enough memory for it — Parallels Desktop is much better and I have configured vagrant box. https://github.com/Parallels/vagrant-parallels/issues/115

So — I just want to open Kinematic and Kinematic should not install VirtualBox/boot2docker — use existing installation (DOCKER_HOST env).

ryansch:
I'd be interested in this to enable our use of vmware fusion.

proudlygeek:
:+1: I really dig this one!

develar:
@ryansch Exactly. Regardless of performance (Vmware is closer, http://www.tekrevue.com/parallels-10-fusion-7-virtualbox-benchmark/), vagrant allows to customize VM more precisely (for example, by default, VirtualBox consumes more than 1GB RAM — vagrant parallels box just restricted to 512 MB and it is enough).

hadronzoo:
:+1:

caseywebdev:
:+1:

tgoeke:
I would vote for supporting VMware Fusion with docker machine.

mainiak:
:+1:

dougborg:
+1 for the ability to specify a driver for docker-machine!

dougborg:
Looks like you can work around this for now by using docker-machine to create and configure a box with whatever driver you like. Just call it `dev` and it looks like Kitematic will just use that one.

After installing kitematic and starting it up for the first time:

```sh
$ docker-machine ls
NAME           ACTIVE   DRIVER         STATE     URL                          SWARM
dev            *        virtualbox     Running   tcp://192.168.99.100:2376
# close kitematic
$ docker-machine rm dev
$ docker-machine ls
NAME           ACTIVE   DRIVER         STATE     URL                          SWARM
$ docker-machine create dev -d vmwarefusion
INFO[0000] Creating SSH key...
INFO[0000] Creating VM...
INFO[0000] Starting dev...
INFO[0000] Waiting for VM to come online...
INFO[0042] "dev" has been created and is now the active machine.
INFO[0042] To point your Docker client at it, run this in your shell: $(docker-machine env dev)
$ docker-machine ls
NAME   ACTIVE   DRIVER         STATE     URL                          SWARM
dev    *        vmwarefusion   Running   tcp://192.168.133.132:2376
# start Kitematic - it will say it is "Starting Docker VM", but it is really just using the one you created:
[...]

PR #1553, opened 2016-03-17
Use existing docker machine if available

zerodivisi0n:
Does not create new VirtualBox docker machine with `default` name if another machine already exists. Uses the first available machine from the `docker-machine ls`. Allows to use machines created with different drivers therefore partially solves the issue #231.


src/utils/SetupUtil.js
```diff
@@ -115,31 +115,50 @@ export default {
...
+          if (machines.length > 0) {
+            // Use first machine for now
+            machine.setName(machines[0].name);
+            exists = true;
+          } else {
+            machine.resetName();
+          }
+        }
+
+        if (!exists) {
```
FrenchBen: Why the 2 `if` cases - these should be combined.

FrenchBen:
Thank you for the PR - Could you make sure to sign your commit, as seen in the contribution guidelines?
https://github.com/docker/docker/blob/master/CONTRIBUTING.md#sign-your-work

Although I like the idea of selecting the first Machine available, this would break some of the existing functionality if the first machine was a remote one. Volume mounting being the first one to come to mind. 
I would filter your 'ls' to look for machines that have a specific driver known to work and local.

src/utils/SetupUtil.js
zerodivisi0n: Wanted to keep checking logic more straightforward. I combined it now.

zerodivisi0n:
Added filter to work only with local drivers. List of drivers I get from [Supported Drivers](https://docs.docker.com/machine/drivers/) and [Available driver plugins](https://github.com/docker/machine/blob/master/docs/AVAILABLE_DRIVER_PLUGINS.md). I don't know how to fully check their workability, so I added only local and active.

Also signed my commits.

FrenchBen:
Finally circling back to this, sorry for the delay.

2 things came to mind:
1. Why save the Machine name? I may create a default after initial launch and can't easily reset it
2. No visual indication of machines - The user has no visual queue that the 'default' machine isn't being used and that some other machine is being loaded. 

Allowing the user to select his/her machine would make more sense IMHO. Although pretty old now, it's what we discussed in the early https://github.com/docker/kitematic/wiki/Machine-Management docs

PR #1603, opened 2016-04-06
Updated path to work properly on linux - Fixes #1591

### LINKED ISSUE #1591: docker cli button on linux does nothing
When you click the "Docker CLI" button in linux, nothing happens, and the dev console reports a TypeError Bad Argument.

A little digging reavealed that on other platforms (not linux) [utils/DockerMachineUtil.js](https://github.com/docker/kitematic/blob/master/src/utils/DockerMachineUtil.js#L163) populates `terminal` with simply a string, and calls `util.execFile([terminal, cmd])`, but on linux the `util.linuxTerminal()` call returns an array.  This is the source of the TypeError, when the true [execFile is actually called](https://github.com/docker/kitematic/blob/master/src/utils/Util.js#L14), 
`child_process.execFile(args[0], args.slice(1), options, (error, stdout) => {....`

`args` contains `[ ['/path/to/terminal', '-e'], 'cmd']`, but it should be `['/path/to/terminal', '-e', 'cmd']`

I'm not sure of the proper way correct this, but below is my patch that fixes it on linux, other platforms untested... hopefully someone can take what I've learned and apply it.

```
diff --git a/src/utils/DockerMachineUtil.js b/src/utils/DockerMachineUtil.js
index 8cd949f..8b3490a 100644
--- a/src/utils/DockerMachineUtil.js
+++ b/src/utils/DockerMachineUtil.js
@@ -160,9 +160,16 @@ var DockerMachine = {
       });
     } else if (util.isNative()) {
       cmd = cmd || process.env.SHELL;
-      var terminal = util.isLinux() ? util.linuxTerminal() : path.join(process.env.RESOURCES_PATH, 'terminal');
+      var terminal;
+      if( util.isLinux() ) {
+          terminal = util.linuxTerminal();
+      }
+      else {
+          terminal = [ path.join(process.env.RESOURCES_PATH, 'terminal') ];
+      }
+      terminal.push(cmd);
       if (terminal) {
-        util.execFile([terminal, cmd]).then(() => {});
+        util.execFile(terminal).then(() => {});
       }
     } else {
       cmd = cmd || process.env.SHELL;
```

FrenchBen:
Updated path to be a consistent array throughout.  Fixes #1591 

### DIFF
--- a/src/utils/DockerMachineUtil.js
+++ b/src/utils/DockerMachineUtil.js
@@ -160,9 +160,10 @@ var DockerMachine = {
       });
     } else if (util.isNative()) {
       cmd = cmd || process.env.SHELL;
-      var terminal = util.isLinux() ? util.linuxTerminal() : path.join(process.env.RESOURCES_PATH, 'terminal');
+      var terminal = util.isLinux() ? util.linuxTerminal() : [path.join(process.env.RESOURCES_PATH, 'terminal')];
+      terminal.push(cmd);
       if (terminal) {
-        util.execFile([terminal, cmd]).then(() => {});
+        util.execFile(terminal).then(() => {});
       }
     } else {
       cmd = cmd || process.env.SHELL;


jeffdm:
LGTM

PR #1698, opened 2016-05-24
Better way to control checkboxes using labels.

sibprogrammer:
Signed-off-by: Alexei Yuzhakov <sibprogrammer@gmail.com>

### DIFF
--- a/src/components/Preferences.react.js
+++ b/src/components/Preferences.react.js
@@ -43,10 +43,10 @@ var Preferences = React.createClass({
           <div className="title">VM Settings</div>
           <div className="option">
             <div className="option-name">
-              Shutdown Linux VM on closing Kitematic
+              <label htmlFor="closeVMOnQuit">Shutdown Linux VM on closing Kitematic</label>
             </div>
             <div className="option-value">
-              <input type="checkbox" checked={this.state.closeVMOnQuit} onChange={this.handleChangeCloseVMOnQuit}/>
+              <input id="closeVMOnQuit" type="checkbox" checked={this.state.closeVMOnQuit} onChange={this.handleChangeCloseVMOnQuit}/>
             </div>
           </div>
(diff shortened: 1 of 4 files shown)


FrenchBen:
I'm not sure what this actually adds?
It seems that you're adding a label to then over-write the label font and make it normal again?

diff
Top is your PR
Bottom is Kitematic 0.10.3

sibprogrammer:
This brings an ability to click on checkbox label instead of trying to click on checkbox itself only. In the other words "Allocate a TTY for this container" also become clickable. It's a common pattern for apps, that not only the checkbox itself is clickable, but the whole label too.

FrenchBen:
Any chance you'd consider doing the same thing for the Preference panel as well?

sibprogrammer:
Ok, I'll check it later today :)

sibprogrammer:
PR has been updated.

FrenchBen:
LGTM

src/components/Preferences.react.js
```diff
@@ -41,14 +41,14 @@ var Preferences = React.createClass({
       vmSettings = (
         <div>
           <div className="title">VM Settings</div>
-          <div className="option">
+          <label className="option">
```
FrenchBen: Anyway to not use an inline element as a block-level element?

sibprogrammer: Yes, it's possible using "label for=...", but it'll be necessary to assign the IDs for input. Is it ok for you?

FrenchBen: Looks better IMHO - Thanks!

FrenchBen:
@jeffdm PTAL

sibprogrammer:
bump :)

FrenchBen:
Thanks for the ping, going to merge it, as it LGTM

Issue #1191, opened 2015-11-06
Missing container ID within Kitematic container info


FrenchBen:
When looking at containers within Kitematic, it would be nice to see the internal container ID as seen by the docker daemon in  `docker ps`.

PR #1702, opened 2016-05-24
Add container ID to sidebar

ClemMakesApps:
Added container ID to sidebar as requested in issue #1191. 
Not sure if this was the desired area to put the information but I'm willing to change it accordingly.
screen shot 2016-05-24 at 4 26 13 pm


FrenchBen:
I would rather have the ID in the 'General' tab - Having it in the sidebar doesn't have any true value. 
Everything you can do with the ID you can do with the name, which kitematic has created for you. 

`docker inspect ff9d` - not very friendly
`docker inspect example2` - is more approachable

ClemMakesApps:
Yeah, that makes sense. So are you saying having it as a non-editable field in the General tab under the Settings tab?

ClemMakesApps:
screen shot 2016-05-25 at 12 36 15 pm

Is this what you were imagining?

FrenchBen:
I would put it after the name and under the `SAVE` button as it isn't something that can be changed.

ClemMakesApps:
screen shot 2016-05-25 at 2 14 59 pm

How about this? The save button spacing looks a little off though. Perhaps it would be wise to have the save button right of the input field? We can also add a copy button right of the ID input field so that it stays balanced and users could use that to copy the container ID.

FrenchBen:
Esthetically it would be better IMHO. Ping @elesant

ClemMakesApps:
Here's a screenshot of what I was trying to communicate:
screen shot 2016-05-27 at 10 01 10 am

- Copy would allow users to copy the container id to their clipboard
- The container ID field is set to the same gray color as the input label to convey to the user that this field is not editable

elesant:
@ClemMakesApps I like this idea. This would work if we move the ID field before NAME, remove the textbox underline on the ID field and remove the border of the COPY button to clearly show that this is a different field than the rest.

ClemMakesApps:
How does this look?
screen shot 2016-05-31 at 9 42 27 am

I also added a feedback message after the COPY button is selected.
screen shot 2016-05-31 at 9 43 05 am

elesant:
This is awesome. Thanks @ClemMakesApps ! +1

FrenchBen:
Merged! Thanks @ClemMakesApps

ClemMakesApps:
No problem. @FrenchBen, you can also close your #1191 issue

FrenchBen:
Interestingly enough, the #1191 was opened for a workshop as a way to show students how to look through issues, implement a solution and create a PR for it ;)

PR #1710, opened 2016-05-27
Fixed container exec

FrenchBen:
Updated the logic for the container exec - Still not perfect but works better in windows when in native.


FrenchBen:
ping @jeffdm

PR #1717, opened 2016-05-28
Fix incorrect syntax that prevents to open the stable version of iTerm.

sibprogrammer:


FrenchBen:
The current syntax is in place to support both pre-2.9 and 2.9+ iTerm.
The original PR can be found here:
https://github.com/docker/kitematic/pull/855

What caused this re-write (if it's just a cleanup, I'm all for it!)

sibprogrammer:
First of all, comparisons like "if version < 2.9 then" does not work at all. Moreover take a precise look at line 52 at https://github.com/docker/kitematic/commit/6c6cefecfcae4e9bd4b79d86ac52e2047cf7346d#diff-52a099fd77c70184568d6beff73aacd7R52
I see very strange comparison operator "≥" and probably it's also a mistake. 

Calls like "create tab with default profile" leads to syntax errors and stops the execution immediately. "Try..catches" doesn't help anyhow against syntax errors.

What's why I think that the author of that PR didn't check the work against iTerm 2.1

Anyways I have an iTerm 2.1.4 and a problem with Kitematic's Exec functionality. It opens in Terminal.app instead of iTerm.

SideshowAlex:
Like @sibprogrammer, I am having problems with iTerm 2.1.4.  I am not an applescript expert, but I do not believe that the script in open_iterm is invalid, and will not compile if only 2.1.4 is installed.

FrenchBen:
I like the simplification and this works quite well for me even when tested against the iTerm 3.0 nightly.
Thanks for the patch 👍

PR #1724, opened 2016-05-30
Changes to avoid the problem when you try to expose the same containe…

alexandrev:
…r port that you have already exposed. Now, when you enter a port value that is already used, the + and the save button gets disabled.


FrenchBen:
Hmm still seems broken for me :(
This is an example of what I'm trying to do:
ports

Issue #1713, opened 2016-05-27
Fix react flattenChildren warnings


ClemMakesApps:
### Expected behavior
No warnings in console

### Actual behavior
![screen shot 2016-05-27 at 10 42 10 am](https://cloud.githubusercontent.com/assets/3363898/15613163/b40a31da-23f7-11e6-9405-65e4c57549f5.png)

### Steps to reproduce the behavior
Run Kitematic and open developer tools console

FrenchBen:
I've seen it creeped it from times, but can't always pin-point where the children with the same key come from. 
Open to suggestions to track it down.

PR #1731, opened 2016-06-01
Fix duplicate react key issue

ClemMakesApps:
This resolves issue #1713. It was caused by duplicate React keys for dynamic elements, so I replaced the existing keys with the following keys (which shouldn't overlap since we do not allow multiple containers with the same name):

`${ContainerName}-${lineNumber}`


FrenchBen:
It's actually difficult for me to replicate even with 5 hello-world-nginx type containers showing. 
I like the simplicity of the fix which should prevent duplicate keys. 

Thanks.

PR #1739, opened 2016-06-04
Only animate the running and restarting icon five times. Continuously…

ruurd:
… animating loads the CPU a lot and we can use the cycles better somewhere else.


FrenchBen:
Sad to see this one being such an issue. LGTM

PR #1781, opened 2016-06-26
Allow empty host port to unbind from PortBindings

keizie:
After random host port set, I couldn't remove it from PortBindings via kitematic
even when I don't need port exposed. (with help of nginx-proxy or something)

This diff allow to inpurt blank host port in Ports tab,
and ignore blank one on setting docker HostConfig.

### DIFF
--- a/src/components/ContainerSettingsPorts.react.js
+++ b/src/components/ContainerSettingsPorts.react.js
@@ -99,7 +99,9 @@ var ContainerSettingsPorts = React.createClass({
       return (i !== key && _.isEqual(v.port, port));
     });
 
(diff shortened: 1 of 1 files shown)


FrenchBen:
Instead of allowing an empty port to unbind, it would be cleaner to allow the port mapping to be removed completely, thus unbinding it.

keizie:
EXPOSEd ports from Dockerfile shall not be allowed to remove.
So empty one is necessary.

FrenchBen:
My fear is that users leave a port blank in order to trigger a 'random port' and are surprised by the result vs expected.

Issue #1779, opened 2016-06-25
Malformed URL on link in "Ports" tab (missing http scheme colon)


andresvia:
### Expected behavior

Click "localhost" link on "Ports" tab, "http://localhost:5000", opens.

### Actual behavior

Click "localhost" link on "Ports" tab, "http//localhost:5000" opens. (no colon : after http)

### Information about the Issue

Bug.

### Steps to reproduce the behavior

  1. Get latest registry image.
  1. Go to "Ports" tab
  1. Change mapped port: container port 5000 to mac port 5000

Fahrradflucht:
Pinging @FrenchBen to close this issue as it is resolved with #1783.

PR #1783, opened 2016-06-27
Remove redundant http:// formatting

ClemMakesApps:
This resolves issue #1779.

The redundant `http://` url formatting when the port is updated is forcing the browser to open URLs with multiple `http://` prefixes. 

For example `http://http://localhost:3000`. The browser parses this url and attempts to load `http//localhost:3000` which fails. 

[http://http://google.com](http://http://google.com) does not equal [http://google.com](http://google.com)


FrenchBen:
LGTM

Issue #1640, opened 2016-04-21
Consider value of `$MACHINE_STORAGE_PATH` 


normenmueller:
### Expected behavior

If `$MACHINE_STORAGE_PATH` is set, then use this one instead of default `~/.docker`.
### Actual behavior

Ignores `$MACHINE_STORAGE_PATH` but creates machines etc. at `~/.docker`.
### Information about the Issue

Cf. [Fix broken --storage-path flag](https://github.com/docker/machine/pull/1941/files)
### Steps to reproduce the behavior
1. Set `$MACHINE_STORAGE_PATH` to a path unequal to `~/.docker`
2. Start Kitematic

FrenchBen:
We actually call out to docker machine directly. I'm guessing the issue is that you set the variable in your ENV, that is unknown to Kitematic. 

Try this.
1. Set your env var for `$MACHINE_STORAGE_PATH`
2. Then run: `open /Applications/Kitematic\ \(Beta\).app`

See if it respects the path.

normenmueller:
Well, we are getting closer :)

```
$ echo $MACHINE_STORAGE_PATH
/Users/nrm/VM/VirtualBox/Docker
$ open /Applications/Docker/Kitematic\ \(Beta\).app
```

Created a new `default` VM, but than complains "Certificate directory does not exist". However:

```
[~/VM/VirtualBox]
$ tree -L 2 Docker
Docker
├── cache
│   └── boot2docker.iso
├── certs
│   ├── ca-key.pem
│   ├── ca.pem
│   ├── cert.pem
│   └── key.pem
└── machines
    ├── bd-hadoop-sandbox
    ├── bd-spark-sandbox
    └── default
```

FrenchBen:
with the env var set, `docker-machine ls` doesn't return the proper machine?

Ping @nathanleclaire Is this the proper way to set this up?

normenmueller:
Still a newbie to docker, which env var must be set in order to get a list of all machines?
The command `eval "$(docker-machine env default)"` is to set env for `default`, right?

nathanleclaire:
It should work fine with just `MACHINE_STORAGE_PATH`.

This is with 0.7.0?

nathanleclaire:
FYI, can't discern if you're doing this here, but creating machines with the same name (even in different storage paths) is likely to cause issues.

normenmueller:
[...]

PR #1831, opened 2016-07-17
Honor DOCKER_CERT_PATH environment in nonNativeSetup

thomasdarimont:
Previously only certificates relative to the home folder
were honored, which made using custom storage locations
via MACHINE_STORAGE_PATH for docker machines quite
cumbersome to use.

This helps with the issue #1640.

### DIFF
--- a/src/utils/DockerUtil.js
+++ b/src/utils/DockerUtil.js
@@ -47,7 +47,7 @@ var DockerUtil = {
         throw new Error('Cannot connect to the Docker daemon. Is the daemon running?');
       }
     } else {
-      let certDir = path.join(util.home(), '.docker/machine/machines/', name);
+      let certDir = process.env.DOCKER_CERT_PATH || path.join(util.home(), '.docker/machine/machines/', name);
       if (!fs.existsSync(certDir)) {
         throw new Error('Certificate directory does not exist');
       }


thomasdarimont:
Just gave this a try with a fresh Windows 10 setup with the following environment variables:
```
MACHINE_STORAGE_PATH=c:\dev\docker\machine
DOCKER_TLS_VERIFY=1
DOCKER_HOST=tcp://192.168.99.100:2376
DOCKER_CERT_PATH=c:\dev\docker\machine\machines\default
DOCKER_MACHINE_NAME=default
DOCKER_TOOLBOX_PATH=C:\Program Files\Docker Toolbox
``` 

With this PR applied Kitematic can create a docker-machine outside of the user home directory.
This enables docker machines to be used by multiple users.

FrenchBen:
Hi @thomasdarimont thank you for the PR - Would you mind signing your commit to confirm with our [Contribution Guidelines](https://github.com/docker/docker/blob/master/CONTRIBUTING.md#sign-your-work)?

Otherwise LGTM! 👍

thomasdarimont:
Thanks @FrenchBe for review.

Yes I comply with your Contribution Guidelines.
I signed-off my commit as described in the Contribution Guidelines.

Cheers,
Thomas

garu57:
Hi all,
i think there are still some problems.
I installed the just downloaded Docker Toolbox on Win 7 Pro SP1.
If i let  the default setup, `%USERPROFILE%\.docker`,  the vm is created once and reused at each Kitematic start.
If i use the setup detailed above, the vm is indeed created in the location pointed by `MACHINE_STORAGE_PATH` and Kitematic starts correctly.

The problem is that **EACH TIME** i start Kitematic the VM is scratched and recreated so loosing any previous work.

Tried also with last released version, 0.13.1, but i got the same behavior.

Regards,
Gabriele

FrenchBen:
@garu57 Do you start kitematic within the PowerShell or with the Desktop shortcut?

garu57:
Launching with start menu shortcut.

FrenchBen:
@garu57 That's why - the Launch menu shortcut will not know about the custom env var. 
You can modify your env vars in windows, so that they are used across the board and that should fix it.

FrenchBen:
@garu57 thanks for providing a bit more detail - When Kitematic starts, it tries to determine if a VM exists and if that VM can be started. If it cannot, then it will create a new one from scratch.
Can you do the following:
1. In Kitematic open the Chrome Dev Tools (CTRL+SHIT+I)
2. Go to the console tab
3. Type the following:
```
var path = require('path')
var fs = require('fs')
var util = require('./utils/Util')
var machine = require('./utils/DockerMachineUtil')
```
4. Provide the output for the following:
* `machine.name()`
* `util.home()`
* `path.join(util.home(), '.docker', 'machine', 'machines', machine.name())`
* `fs.existsSync(path.join(util.home(), '.docker', 'machine', 'machines', machine.name()))`

Thanks

garu57:
@FrenchBen I really had a hard time locating the SHIT key. :grinning:
Anyhow, on my keyboard, it is the ALT key. 
Don't ask me why but neither CTRL+SHIFT+i nor F12 do anything. Only CTRL+ALT+i opens the Chrome Dev Tools.

After starting Kitematic you see that the VM is created in the location pointed by env vars:
```
C:\home\VirtualBoxVMs\Docker>tree /f
Folder PATH listing for volume OS
Volume serial number is DEB5-F005
C:.
├───cache
│       boot2docker.iso
│
├───certs
│       ca-key.pem
│       ca.pem
│       cert.pem
│       key.pem
│
└───machines
    └───default
        │   boot2docker.iso
        │   ca.pem
        │   cert.pem
        │   config.json
        │   disk.vmdk
        │   id_rsa
        │   id_rsa.pub
        │   key.pem
        │   server-key.pem
        │   server.pem
        │
        └───default
            │   default.vbox
            │   default.vbox-prev
            │
            └───Logs
                    VBox.log
                    VBoxHardening.log
```

But it seems that Kitematic is a bit confused, it still looks for VM in the default location.

image

Cheers, 
Gabriele

FrenchBen:
@garu57 that's what I was afraid of and why Kitematic can't seem to find the VM, because it doesn't see your User specific env vars.

Can you try to add them within the System Env Vars instead?
System env vars

Otherwise you may need to add them to the shortcut :(

garu57:
@FrenchBen I do not agree that Kitematic doesn't see the user env vars.
If it were so why it creates the VM in the location pointed by `MACHINE_STORAGE_PATH` var and it is able to find the certificates in the location pointed by `DOCKER_CERT_PATH` var, and it is able to start the VM in the location pointed by `MACHINE_STORAGE_PATH` var and even install containers in it?
The problem is that the startup check for machine existence ignores the `MACHINE_STORAGE_PATH` var and it is done in the default location.

Anyhow, even if once the vars are in the environment it doesn't matter if they come from "User Variables" or "System variables", i did what you asked for.
I put the vars into "System variables".
image
```
[HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Session Manager\Environment]
...
"DOCKER_CERT_PATH"="C:\\home\\VirtualBoxVMs\\Docker\\certs"
"DOCKER_MACHINE_NAME"="default"
"DOCKER_TLS_VERIFY"="1"
"DOCKER_TOOLBOX_INSTALL_PATH"="C:\\Program Files\\Docker Toolbox"
...
"MACHINE_STORAGE_PATH"="C:\\home\\VirtualBoxVMs\\Docker"
...
```
and here is the result:
image

Again Kitematic ignores `MACHINE_STORAGE_PATH` var during startup check and it scratches the VM it previously created in the location pointed by `MACHINE_STORAGE_PATH` and recreates a new VM in the location pointed by `MACHINE_STORAGE_PATH` var.

PR #1833, opened 2016-07-17
append support external hub

bessdsv:
appended possibility pull image from others docker hubs.
because i am new in ecmascript 6 and reactjs, this code not final.
I wait some comments from project developers


FrenchBen:
Changing the hub endpoint isn't the only change needed - There needs to be an authentication mechanism in place to allow users to authenticate against that registry. 
How is this handled here?

PR #1559, opened 2016-03-20
Fix for menu bar not displaying in windows

Bigguy34:
So I spent a couple days on this, and there are a couple problems. Firstly the main reason that the menu bar doesn't display in windows is because kitematic sets `frame:false` in the settings of the browser window. This prevents it from being displayed in Windows. So I implemented a fix. By removing the `frame:true` and changing it to `frame: os.platform() === 'win32'`, stuff started working. But there are a lot of UI problems. I tried to fix most of them. Here is a screen shot of what I have so far(And yes hiding the bar is supported with the `alt` key). 
menubar

Here is window before the alt key is pressed.
withoutmenubar

The biggest problem is that I had to get rid of the header with the login and logo. I plan on putting back the information, I just want to make sure first that this is worth pursuing. If it is I would be curious to hear from some of the maintainers where the best place to put the rest of this information(I was thinking of putting it in top menu bar, or moving it to the bottom some where). I would be curious to here some thoughts. Thanks!

PR #1835, opened 2016-07-18
fix: replace custom window buttons with native

### LINKED ISSUE #1816: Option-Click on Mac Should Maximize
### Expected behavior
Option-clicking on the green button in the title bar should cause the window to maximize without going fullscreen.

### Actual behavior
The window goes into fullscreen mode.

  1. Option-click on the green button in the titlebar.

mikaoelitiana:
fixes #1816

### DIFF
--- a/src/browser.js
+++ b/src/browser.js
@@ -31,7 +31,7 @@ app.on('ready', function () {
     'min-height': os.platform() === 'win32' ? 260 : 500,
     'standard-window': false,
     resizable: true,
-    frame: false,
+    frame: true,
     show: false
   });
 
--- a/src/components/Header.react.js
+++ b/src/components/Header.react.js
@@ -104,27 +104,6 @@ var Header = React.createClass({
       </div>
     );
   },
-  renderWindowButtons: function () {
-    let buttons;
-    if (util.isWindows()) {
-      buttons = (
-        <div className="windows-buttons">
-        <div className="windows-button button-minimize enabled" onClick={this.handleMinimize}><div className="icon"></div></div>
-        <div className={`windows-button ${this.state.fullscreen ? 'button-fullscreenclose' : 'button-fullscreen'} enabled`} onClick={this.handleFullscreen}><div className="icon"></div></div>
-        <div className="windows-button button-close enabled" onClick={this.handleClose}></div>
-        </div>
-      );
-    } else {
-      buttons = (
-        <div className="buttons">
-        <div className="button button-close enabled" onClick={this.handleClose}></div>
-        <div className="button button-minimize enabled" onClick={this.handleMinimize}></div>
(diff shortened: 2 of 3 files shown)


FrenchBen:
Do you think you can fix the layout for windows, so that it's consistent with Linux+OSX. 
Login button on the left, kitematic logo on the right?
Like:

FrenchBen:
@mikaoelitiana In case you missed it, there were some nice enhancements here as well: https://github.com/docker/kitematic/pull/1559

mikaoelitiana:
Oh that's great!! I think they have done good job in this mentionned PR so maybe we can close this one

mikaoelitiana:
I will check the layout in Windows and PR if any change needed there.

FrenchBen:
A couple of things weren't smooth in the mentioned PR and unfortunately the original dev wasn't able to finish it :(

mikaoelitiana:
I have some trouble testing this in Windows as I have no Windows computer and tried to launch it in a VM. Docker won't be installed in the VM as Windows test VMs don't have 64bit version. 
Can someone check this out for me in Windows?

FrenchBen:
I'll give it a shot - thanks for the help

Bigguy34:
Whoa... Does this PR get rid of the frameless option? Sorry @FrenchBen changing jobs, and buying a new house really cuts down on my free time. If we are going to get rid of that electron config option everywhere(Linux, OSX, Windows) my original PR will work really well, and it is much easier. I think for now this will be an easier direction than trying to implement a native looking menu in React. I can get my original PR up to speed if need be.

FrenchBen:
@Bigguy34 It may be a lot easier to do so - the original intent was to keep a similar look and feel across all OS, but seems to add some un-needed overhead. WDYT?

Bigguy34:
Yea so in my PR I had it optionally switching between frame:true and frame:false depending on whether or not a user was in windows or mac. In this PR we just set frame:true everywhere, and just handle the position of elements in the topbar based on whether or not we are in windows. I do think the work I did around the menu would definitely still apply. Originally I was going to try and remake all the "native" looking controls in React, [At a high level something like this](http://stackoverflow.com/questions/35876939/frameless-window-with-controls-in-electron-windows). If your ok with it we add the frame back in everywhere, we will just have to work a little harder to make it look pretty in windows. WDYT @FrenchBen I am working on getting my branch up to date, and I think I will finally have time to finish my PR. I am really sorry it to so long. What are your thoughts @mikaoelitiana ?

mikaoelitiana:
@Bigguy34 I am OK with the idea to continue your PR which is already quite advanced and close this one

FrenchBen:
Closing in favor of @Bigguy34's PR #1559

PR #1847, opened 2016-07-24
Electron debian fix

sbacic:
This fixes the lintian linter errors for the debian package so that it can be installed from the .deb package without tripping the linter. You can check that it passes the linting with 
`lintian Kitematic_0.12.1_amd64.deb `. 

I had to resort to overriding most of the linting messages because they were related either to the way electron packager does the packaging or the Kitematic's module dependencies, both of which are out of scope of a localized fix. You can see the 14 error messages with `lintian --display-level=">=important" -o  Kitematic_0.12.1_amd64.deb`. Most have to do with the embedded libraries.

I've left a comment for each override in the first commit and removed them and collapsed the overrides in the second commit to keep everything clean. 

This bit is kind of **important** - I haven't been able to find a mailing address for Kitematic so I used yours for the maintainer field. 

To finish off, the package should now work properly in Debian, though some things, such as the permission fixes and moving the non-JS libraries out of the package and leaving them as external dependencies should be considered for the future.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -317,15 +317,36 @@ module.exports = function (grunt) {
     },
     'electron-installer-debian': {
       options: {
-        productName: LINUX_APPNAME,
+        name: BASENAME.toLowerCase(), // spaces and brackets cause linting errors
+        productName: LINUX_APPNAME.toLowerCase(),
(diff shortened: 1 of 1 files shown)


FrenchBen:
LGTM thanks for updating the package builder

PR #1878, opened 2016-08-02
Folder limit

### LINKED ISSUE #1707: Docker for Windows/Mac should allow sharing of non-User volume (e.g. C:/tmp)
### Expected behavior
With docker for Windows, you can share drives like C: in settings. So it should be possible to use something like 

docker run -d -v c:/tmp/Gastrobase:/home/shiny/Gastrobase ....

In fact, some user recommend NOT to use the users folder any more because of user registration problems (not confirmed)

### Actual behavior
In Kitematic, you get an error message when trying to change the local folder to c:/tmp or similar

FrenchBen:
* Removed folder limitation for native apps
* Add some caching to the requests
* Added preference of VM vs Native 

Fixes #1707

this can't be merged until this PR is merged: 
https://github.com/alltherooms/cached-request/pull/12

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "Kitematic",
-  "version": "0.12.0",
+  "version": "0.12.0.1",
   "author": "Kitematic",
   "description": "Simple Docker Container management for Mac OS X, Windows and Ubuntu.",
   "homepage": "https://kitematic.com/",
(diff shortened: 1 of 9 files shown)


dluc:
Nice add, as mentioned in #1885, perhaps you could allow the user to define the order: e.g. "probe VirtualBox first, then Docker" or viceversa.
Thanks!

FrenchBen:
That order wouldn't work well, as the normal behavior for Kitematic is to create the VM if it doesn't exist.

PR #1960, opened 2016-09-01
Add engines entry to package.json

### LINKED ISSUE #1944: Specify required node version
[CONTRIBUTING.md](https://github.com/docker/kitematic/blob/master/CONTRIBUTING.md) only says that Node is needed but it does not say which version should be used. Correct me if I'm wrong but this suggests to me that the current version should be good. Especially because there is neither an .nvmrc file in the project nor an `engines` key set in package.json.

Turns out the tests won't even run on node 6.

Looking in [.travis.yml](https://github.com/docker/kitematic/blob/master/.travis.yml) I see that the CI tests run on 4.2.2 what suggests that this is the version everyone should develop against.

Options of fixing that are:
- Adding the version to CONTRIBUTING.md (Which has a high risk of going stale I think)
- Adding a [.nvmrc](https://github.com/creationix/nvm#nvmrc)
- Add an `engines` key to package.json

Fahrradflucht:
Fixes #1944

This adds an engines entry to package.json. It's set to 4.1.1 which is the node version used in all electron versions between [0.33.2](https://github.com/electron/electron/releases/tag/v0.33.2) and [0.36.0](https://github.com/electron/electron/releases/tag/v0.36.0) what covers the current ([0.35.4](https://github.com/docker/kitematic/blob/master/package.json#L23)).

--

This actually brings up the question why the tests on travis [run against 4.2.2](https://github.com/docker/kitematic/blob/master/.travis.yml#L5) as that isn't the node version kitematic actually runs on doesn't it?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: false
 
 language: node_js
 node_js:
-    - "4.2.2"
+    - "4.1.1"
 
 cache:
   directories:
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
   },
   "license": "Apache-2.0",
   "electron-version": "0.35.4",
+  "engines" : { "node" : "4.1.1" },
   "dependencies": {
     "alt": "^0.16.2",
     "ansi-to-html": "0.3.0",


Fahrradflucht:
I understand that but I don't get why. Am I wrong in thinking that 0.35.4 runs on 4.1.1?

FrenchBen:
It does - But 4.2.2 is considered LTS and thus should be maintained for an extended period, while still getting but fixes/patches.
https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V4.md#4.2.2

From Node:

Changes in an LTS-covered major version are limited to:
1. Bug fixes;
2. Security updates;
3. Non-semver-major npm updates;
4. Relevant documentation updates;
5. Certain performance improvements where the risk of breaking existing applications is minimal;
6. Changes that introduce large amount of code churn where the risk of breaking existing applications is low and where the change in question may significantly ease the ability to backport future changes due to the reduction in diff noise.

Fahrradflucht:
I will change it if that's your opinion but I don't agree. Sure it would be great to run on the LTS since it's still supported but just changing the version in travis and the engines key doesn't change anything on the fact that Kitematic runs on 4.1.1 and that tests that pass on 4.2.2 could have different results in the application even if the risk is relatively low. 

The only way to truly run on a supported node version is to upgrade the used Electron version.

--
 Anyway, I will push an update.

pmario:
>It does - But 4.2.2 is considered LTS and thus should be maintained for an extended period, while still getting but fixes/patches, etc.

I'm confused. IMO it's 4.x that's considered to be LTS. So all the fixes up to 4.5.0 are missing. .. So for me testing against 4.2.2 is not a real argument any more.

Fahrradflucht:
Good point. I honestly didn't check if 4.2.2 is the current version of the 4 LTS branch but if it isn't then that does make less sense then ever to me. Kitematic runs on 4.1.1 and the current version is 4.5 whats the point in running CI against and develop on 4.2.2?

FrenchBen:
@pmario It's definitely due for an update to `4.5.0` - I'm open to discuss what should be our minimum vs recommended version.
@pmario @Fahrradflucht WDYT?

Fahrradflucht:
My opinion is still that 4.1.1 should be the version that the development and the CI should be on - no ranges, no min no max. It's the version that Kitematic uses and everything that isn't that does open the possibility of passing tests but failing in production. Even if the changes are non-breaking, noticing existing bugs that are present in production is valueable in development and testing.

pmario:
I just had a look at electron.atom.io. It seems they don't have anything LTS related. They are bumping the minor version number whenever they change node version and chrome version. ... Just curious. Is the API stable?

FrenchBen:
@pmario tbh I have no clue - If we're all happy keeping it at 4.1.1 then let's just do that. 
Upgrading to the latest electron is a bit more challenging.

Fahrradflucht:
Okay I change it back. Should I change CI as well ? (travis)

FrenchBen:
Yes please do if you don't mind so that everything is in sync! Thanks for the help @Fahrradflucht

Fahrradflucht:
@FrenchBen Done!

FrenchBen:
LGTM - Thanks!

I'll try to go through and upgrade us to the latest `4.x`, since it's still the recommended install:
https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions

Fahrradflucht:
Yeah that would be the case if this was a regular node project but since it's an Electron app it makes not much sense to track the recommended node versions.

Electron tracks the node version that is nearest the v8 version in the used Chrome and goes through a lot of hassle to make that happen (including patching chrome I think).

In [`v.0.36.0`](https://github.com/electron/electron/releases/tag/v0.36.0) they upgraded to `5.1.1` and the current (`1.3.5`) uses node`6.3.0` which maps to Chromium `52.0.2743.82` which is V8 `5.2.361.43`.

In my opinion it makes not much sense for the reason we discussed early in this thread to track node versions apart from electron. Instead the focus should be on the electron version and the node version that comes with the electron version.

To answer @pmario's [question](https://github.com/docker/kitematic/pull/1960#issuecomment-244761963). The electron API is stable since [`1.0.0`](https://github.com/electron/electron/releases/tag/v1.0.0). That means that ones Kitematic hits updates to a Electron `>1.0` it will be much easier to track the latest version (until v2 comes along of course).

Issue #1455, opened 2016-02-08
Feature: a way to disable the web preview or just display a screenshot instead


adrian7:
I'm using Kitematic 0.9.3 on Mac OS X.

I have a project (my bad) which does a lot of javascript. Unfortunately this causes Kitematic UI to freeze until the website loads in web preview. 
I would like to be able to disable web preview or have that component display a screenshot instead of loading the page and executing all the scripts.

Regards,
Adrian.

FrenchBen:
Thanks for the info - Would  you be able to write-up how you see it working?
This would help anyone willing to work on this feature.

adrian7:
I think ideally in the View menu, have an item 

Web preview 
 - Browser
 - Screenshot only
 - Disable

alexandrev:
@FrenchBen, what do you think about the approach? 3 options? Only a check to disable it? If you want I could go for it if you think this is important..

FrenchBen:
@alexandrev I don't think this is important, as this is the first use case - I would like the OP to take a stab at it, if he thinks this is something important to him.
I'm sure it's a nice feature, but there's a bit more to it due to the info for that container no longer being saved in the container data but rather in somewhere in kitematic.

ericmann:
Chiming in here because Kitematic's built-in browser view gave me a bit of a headache today. It's not that I'm running a container that prints a lot of JS, it's that my container is attempting to parse specific traffic sent to a TCP port it monitors. Normally, everything's fine. But if I alt-tab to the Kitematic window, my app will crash because of the flood of _web_ traffic it gets from the browser attempting to hit the open port directly.

Being able to turn this off would be a huge win for me.

FrenchBen:
@ericmann As explained above, I'm waiting for someone to step up and provide details on how this feature will be implemented.
Currently we're focused on bug fixes rather features like this. Any help is appreciated however.

bostonaholic:
I'd like to reopen this feature exploration with PR #2016

FrenchBen:
@bostonaholic Since you're actually moving on the PR, I'm re-opening. Thanks for carrying this.

jpduckwo:
my kitematic crashes every time I open it and click on my nginix container which hosts an angular js app. I would also like to disable the web preview which I think is causing the crash.

scytacki:
[...]

PR #2016, opened 2016-09-25
click 'X' to close a preview

bostonaholic:
This is a PR to re-open the conversation about #1455.

Things to consider for this feature:
1. Expand `.left` when all of the `.right` windows have been closed.
   
   a. This will allow the user to close the panels on the right side and expand the logs on the left. This is great for users with smaller screens that want to expand the logs that may be wrapping.
2. Add a way to re-open the panels on the `.right` instead of just close.
   
   a. Maybe a collapse/expand instead?
   
   b. As of now, the workaround is to click to another container and back again. The closed panel will reappear.
3. Save the open/closed state for the panels.
   
   a. Once we have a way to reopen a panel (above), save the open/closed state such that moving away from the container and back, will not show the panel again.

Signed-off-by: Matthew Boston matthew@matthewboston.com

### DIFF
--- a/src/components/ContainerHomeFolders.react.js
+++ b/src/components/ContainerHomeFolders.react.js
@@ -12,9 +12,18 @@ const dialog = remote.dialog;
 import mkdirp from 'mkdirp';
 
 var ContainerHomeFolder = React.createClass({
+  getInitialState: function () {
+    return {display: true};
+  },
+
   contextTypes: {
     router: React.PropTypes.func
   },
+
+  handleClickClose: function () {
+    this.setState({display: false});
+  },
+
   handleClickFolder: function (source, destination) {
     metrics.track('Opened Volume Directory', {
       from: 'home'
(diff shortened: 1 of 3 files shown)


FrenchBen:
It would be nice if this could be made into a slide in/out menu perhaps, to allow the ability to close and re-open it. 
If you look at xcode, they have a nice interface to designate the panels that get shown. 

WDYT?

bostonaholic:
@FrenchBen I think that's a great idea. It's what I was thinking about in 2 and 3 above. I haven't had a lot of time lately to finish this, but I'll be back on it this week.

FrenchBen:
No problem - Time is always the enemy. Thank you for the help!

FrenchBen:
ping @bostonaholic

bostonaholic:
@FrenchBen Sorry for the delay. I'm still trying to figure out how I want the slide in/out to work and seeing if it's best to write it from scratch or if there's a react/vanilla.js library that would help.

Do you have any suggestions for how you would want this to look and how it behaves?

FrenchBen:
tbh I've never really been bothered by the preview. 
There are however a couple of different modes possible, since the preview pane may also include the port links.
Perhaps instead of a slide left/right, it should be an accordion style? (up/down)
Something like: http://react-component.github.io/collapse/examples/simple.html 

WDYT?

bostonaholic:
The reason the left/right is the way to go, is to allow the logs on the left to expand. Sometimes on a smaller screen, my logs will wrap, making it difficult to read.

FrenchBen:
@bostonaholic what about adapting the drawer component?
http://react-component.github.io/drawer/examples/simple.html

bostonaholic:
Cool. Something like that could work.

bostonaholic:
Closing to clear out my stale PRs. Feel free to reopen if the need arises.

PR #2042, opened 2016-10-11
Kitematic does not run on Linux distributions

rafaelstz:


FrenchBen:
@rafaelstz Where does it not run? We have a debian package that is released:
https://github.com/docker/kitematic/releases/tag/v0.12.0

Issue #1648, opened 2016-04-26
Kitematic Linux Exec Terminal


ryanleesipes:
Hey guys, using Kitematic on Arch. Really able to get things working pretty well except for the Exec button. Upon clicking I get the following error:

Unhandled rejection Error: Encountered an error: Error: Command failed: /bin/sh -c /usr/bin/x-terminal-emulator -e docker exec -it ubuntu-2 sh
Unknown option -e

    at /opt/Kitematic (Alpha)-linux-x64/resources/app.asar/utils/Util.js:48:18
    at ChildProcess.exithandler (child_process.js:232:5)
    at emitTwo (events.js:87:13)
    at ChildProcess.emit (events.js:172:7)
    at maybeClose (internal/child_process.js:818:16)
    at Socket.<anonymous> (internal/child_process.js:319:11)
    at emitOne (events.js:77:13)
    at Socket.emit (events.js:169:7)
    at Pipe._onclose (net.js:469:12)

Screenshot Attached.

Am I missing a dependency? Just need some direction. @zedtux perhaps you have some insight here?
![screenshot from 2016-04-26 06-35-36](https://cloud.githubusercontent.com/assets/2141621/14816897/1f5946a2-0b79-11e6-8d2e-7755122465ac.png)

zedtux:
@ryanleesipes can you post the output of the command `x-terminal-emulator --help` and also `lsb_release -a`.

ryanleesipes:
I'm trying to use terminix as the terminal. But your question prompted me to check if it had a '-e' argument, and of course it doesn't. It features a '-x' instead. Perhaps in a future release we can pop in a way to change this flag?

I can use the command with xterm, but even gnome-terminal throws up an error:

Unhandled rejection Error: Encountered an error: Error: Command failed: /bin/sh -c /usr/bin/x-terminal-emulator -e docker exec -it mycroft-dev sh
Failed to parse arguments: Unknown option -it

Finally, I've noticed the default command to run is 'sh', it would be cool if this could be toggled to bash if the user so pleased.

zedtux:
What Linux distro are you using?

ryanleesipes:
Heh, Arch. So I'm on a moving target.

bebehei:
@zedtux Thanks for the good work. I'm astonished, someone took the effort to port such a nice looking app onto linux.

This issue shows a small corner case of the differences between Ubuntu and Arch.

[...]

PR #2084, opened 2016-10-31
improve linux terminal link error message

bebehei:
After clicking on the `exec`-Button, a message popped up, stating `The terminal emulator symbolic link doesn\'t exists. Please read the Wiki at https://github.com/docker/kitematic/wiki/Early-Linux-Support.`.

Sadly:
1.  the message gives no indication of the problem.
2.  The link given in the message, points to the wiki-site [Early-Linux-Support](https://github.com/docker/kitematic/wiki/Early-Linux-Support), which is except a link is empty
3.  The Link contains an anchor [#early-linux-support-from-zedtux](https://github.com/kitematic/kitematic/wiki/Common-Issues-and-Fixes#early-linux-support-from-zedtux), which does not exist on the linked  Wiki-Site.

I got my indication out of #1648.

It may also be good to remove the bad links in the message (and) remove the bad wiki-pages.

### DIFF
--- a/src/utils/Util.js
+++ b/src/utils/Util.js
@@ -207,7 +207,7 @@ module.exports = {
       dialog.showMessageBox({
         type: 'warning',
         buttons: ['OK'],
-        message: 'The terminal emulator symbolic link doesn\'t exists. Please read the Wiki at https://github.com/docker/kitematic/wiki/Early-Linux-Support.'
+        message: 'The symbolic link /usr/bin/x-terminal-emulator does not exist. Please read the Wiki at https://github.com/docker/kitematic/wiki/Early-Linux-Support for more information.'
       });
       return false;
     }


FrenchBen:
Thanks for fixing this - Would you be able to provide an alternative terminal emulator that worked for you?

bebehei:
Using the symlink `/usr/bin/x-terminal-emulator` is the perfect thing. It gives programmers the ability to use any terminal-emulator without big hassle.

It's just not linked on every system. This PR only improves the error message, nothing more. The next user should know, what to fix.

FrenchBen:
It seems that the wiki should then be updated, as the page you mention: https://github.com/docker/kitematic/wiki/Early-Linux-Support points to another wiki page, which has none of the Linux information.

Do you mind updating that page also noting what distro needs the symlink created?

bebehei:
@FrenchBen I changed [the corresponding wiki page](https://github.com/docker/kitematic/wiki/Early-Linux-Support). Could you please merge it?

FrenchBen:
Thanks for the patch!

PR #2313, opened 2017-01-31
Add tab with configuring container networks

### LINKED ISSUE #1249: Feature proposal: Add Networking to containers
This proposal would keep networking to a minimal for now, but with a big impact.
The idea is as follow:
- [ ] Enable network support within Kitematic Preferences (this adds a new 'kitematic' networ)
- [ ] Add network to containers (with networking enabled, this would show the network that the container can attach to)

Keep in mind that the above is a rough draft and comments are welcome as to how users would like this to work. 
I would like to keep this as a small feature for now which can then grow later on.  Adding the default 'kitematic' network would allow for all things to talk to each other on a common network.

megazoll:
These PR add 2 features:
1. Saving network settings when container updates.
2. Configuring container network on separate tab in container settings.

Fixes #1249 

Signed-off-by: Vyacheslav Salakhutdinov <megazoll@gmail.com>

### DIFF
--- a/src/utils/DockerUtil.js
+++ b/src/utils/DockerUtil.js
@@ -9,6 +9,8 @@ import hubUtil from './HubUtil';
 import metrics from '../utils/MetricsUtil';
 import containerServerActions from '../actions/ContainerServerActions';
 import imageServerActions from '../actions/ImageServerActions';
+import networkActions from '../actions/NetworkActions';
+import networkStore from '../stores/NetworkStore';
 import Promise from 'bluebird';
 import rimraf from 'rimraf';
 import stream from 'stream';
@@ -85,6 +87,7 @@ var DockerUtil = {
     this.placeholders = JSON.parse(localStorage.getItem('placeholders')) || {};
     this.refresh();
     this.listen();
+    this.fetchAllNetworks();
 
     // Resume pulling containers that were previously being pulled
     _.each(_.values(this.placeholders), container => {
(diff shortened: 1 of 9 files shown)


src/utils/DockerUtil.js
```diff
@@ -159,6 +160,12 @@ var DockerUtil = {
         containerData.HostConfig.PublishAllPorts = true;
       }
 
+      containerData.NetworkingConfig = {
+        EndpointsConfig: _.mapObject(containerData.NetworkSettings.Networks, (network) => {
```
FrenchBen: Unfortunately this doesn't always work - If no-network has been set, you're creating the container for the first time, you'll end up with:
`TypeError: Cannot read property 'Networks' of undefined`

src/utils/DockerUtil.js
```diff
@@ -263,6 +271,68 @@ var DockerUtil = {
     });
   },
 
+  fetchAllNetworks () {
+    this.client.listNetworks((err, networks) => {
+      if (err) {
+        networkActions.error(err)
+      } else {
+        networks = networks.filter(n => n.Name !== 'host').sort((n1, n2) => {
```
FrenchBen: Why is `host` network filtered out here? Granted for 99% of the cases it's best to avoid it

megazoll: When I trying to connect any container to `host` network from cli i got an error:

```
$ docker network connect host blackfire
Error response from daemon: Container cannot be disconnected from host network or connected to host network
```

src/utils/DockerUtil.js
megazoll: Also this will not work with multiple networks (https://github.com/docker/docker/issues/29265), I will trying to solve it.

src/utils/DockerUtil.js
FrenchBen: By default the container will be attached to the bridge network. You cannot have both the host and some other network, but can create a container with the host network:
```
docker run --net host --rm -it alpine sh
```

megazoll:
I added listener and ability to create container with selected network, please review.

PR #2397, opened 2017-02-28
Update ContainerDetailsSubheader.react.js

kenneyhe:
process environment will allow user to see private docker hub


kenneyhe:
redundant from to change

PR #2398, opened 2017-02-28
Use environments to determine hub to pull and login

kenneyhe:
Add environment to change default hub.docker.com


kenneyhe:
still need to merge one more

PR #2476, opened 2017-03-19
Terminal shell

daaru00:
Added setting into preference to select which shell you want to exec using "Exec" button.
preview

### DIFF
--- a/src/components/AccountLogin.react.js
+++ b/src/components/AccountLogin.react.js
@@ -28,7 +28,7 @@ module.exports = React.createClass({
   validate: function () {
     let errors = {};
 
-    if (this.state.username.indexOf('@') > -1 && !validator.isEmail(this.state.username) {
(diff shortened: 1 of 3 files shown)


FrenchBen:
@daaru00 Thanks for putting work into making the Shell selection an option. 
Could you Sign your commits? see: https://github.com/docker/docker/blob/master/CONTRIBUTING.md#sign-your-work

For windows containers, we may need to add the `powershell` option as well.

daaru00:
> Could you Sign your commits? see: https://github.com/docker/docker/blob/master/CONTRIBUTING.md#sign-your-work

Sorry, I didn't see it. I have to make a commit amend?

> For windows containers, we may need to add the powershell option as well.

Right! but is hard to select the correct Shell based on image OS. 
Maybe into a container's settings but could be frustrating specify it on each one container.

FrenchBen:
I believe you should be able to do: `git commit --amend -s` and then push again

PR #2509, opened 2017-03-27
Added hostname settings to container

daaru00:
According to https://github.com/docker/kitematic/issues/2350#issuecomment-287486749 I edited ports tab adding container's hostname configurations:
preview
If not set the current container hostname is loaded.

### DIFF
--- a/src/components/ContainerSettings.react.js
+++ b/src/components/ContainerSettings.react.js
@@ -47,7 +47,7 @@ var ContainerSettings = React.createClass({
               </Router.Link>
               <Router.Link to="containerSettingsPorts" params={{name: container.Name}}>
                 <li>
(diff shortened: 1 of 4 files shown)


daaru00:
It may be considered to add an input validation or an error handler.

daaru00:
Something like this: https://github.com/daaru00/kitematic/commit/b8a40f77e7188ed53ace2e6af214b12534c851d6

hostname_not_valid

FrenchBen:
@daaru00 that would be cool as well - If someone sets the wrong hostname, it'll break the current container and thus may help for inexperienced users.

daaru00:
@FrenchBen exactly! Can you test the commit changes?

PR #2545, opened 2017-04-03
Added pointer cursor to clickable elements

koole:
I've added pointer cursors to some of the clickable elements in the application. This is to improve usability of the application, because for some of the elements there was no indication at all they were clickable.

For the `.more-menu` element next to the _create_ button I also added a hover state to make it even more clear that this is a clickable element .

### DIFF
--- a/styles/left-panel.less
+++ b/styles/left-panel.less
@@ -70,7 +70,7 @@
       a {
         color: inherit;
         flex-shrink: 0;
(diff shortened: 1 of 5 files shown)


FrenchBen:
I may be missing something, but the left-side panels don't have any `pointer` for me, although they are clickable items. Can you add them too?

PR #2556, opened 2017-04-09
Added links section to network container tab

daaru00:
List linked containers and open container page.
Allow to manage container links adding new or remove an existing existing one. 
Add style to select input.
links_restyle


FrenchBen:
@daaru00 you're moving fast on the PRs 💟  👍 

Unfortunately links are somewhat of a legacy concept; could we perhaps show the list of containers on the same network instead?

daaru00:
With latest commit when container start all linked container will be automatic start (if the new option in preference is set)
preference-linked

PR #2574, opened 2017-04-15
Fix login with email in AccountLogin.react.js

### LINKED ISSUE #2571: Can't login to docker hub windows
I can't seem to login with a valid email
capture

nicolasbrugneaux:
If the email is correct, then the `else if` would evaluate to false.
Fix it by nesting the email check _only_ if a `@` is detected.


FrenchBen:
thanks @nicolasbrugneaux for the quick patch!

PR #3216, opened 2017-10-26
Use local instead of local-library for local images namespace

### LINKED ISSUE #2870: Bug: Tagged image pulled from shell results in 404 when creating new container from Kitematic
### Expected behavior

OSX 10.12.5
Kitematic 0.17.1
Docker version 17.06.0-ce, build 02c1d87

after pulling ruby:2.3, from the shell, I see the image listed when I issue the command `docker images` from the shell, as well as My Images in Kitematic. When I click "Create" within Kitematic My Images view, I expect a container to be launched.

### Actual behavior

I immediately get an error: 

> (HTTP code 404) no such image - no such image: local-library/ruby:2.3: No such image: local-library/ruby:2.3

### Information about the Issue

I have other images I pulled using the shell, and I'm able to launch them from Kitematic, but this is the only one that specified a tag when I pulled the image, so possibly relevant. I am also able to run the local image from the shell with `docker run -it ruby:2.3` and get expected behavior.

  1. from bash shell, execute: `docker pull ruby:2.3`
  2. In Kitematic, click + New to add a new container. In the following view, select "My Images" tab, then for Ruby 2.3 image, select "Create"

mhelmetag:
Fixes #2870 

I think this was just a matter of the namespace being `local-library` instead of `local`.

issue_2870

**Note:** gif is a little long but should be good for proof that the change works.

### DIFF
--- a/src/utils/DockerUtil.js
+++ b/src/utils/DockerUtil.js
@@ -291,7 +291,7 @@ var DockerUtil = {
             repo = imageSplit[0]+':'+imageSplit[1];
           }
           if (repo.indexOf('/') === -1) {
-            repo = 'local-library/' + repo;
+            repo = 'local/' + repo;
           }
           [list[idx].namespace, list[idx].name] = repo.split('/');
         });


mhelmetag:
Thanks!

PR #3265, opened 2017-11-06
Update eslint and eslintrc to not get sad when parsing async

mhelmetag:
I was having a lot of trouble getting eslint to parse all the files in `src`. It seemed to be sad about the `async` keyword and then, after updating eslint, also `import` and `export`. This seems to have made it better.

I also would like to ask if people would be interested in converting the `.eslintrc` from YAML to JSON. I feel like the eslint docs don't really have YAML examples... which, IMO, can make config changes harder.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,11 +3,9 @@ root: true
 plugins:
     - react
 
-ecmaFeatures:
-    modules: true
-    jsx: true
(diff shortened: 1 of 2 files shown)


FrenchBen:
I liked the simplicity of YAML but if JSON is preferred I don't mind switching to it.

PR #3313, opened 2017-11-16
Bumped electron version, and fixed some minor issues with that.

alex4o:
Replaced electron-prebuild with electron
Where there was shell useg replaced
`import shell from 'shell'`
with
`import {shell} from 'electron'`
Also minor style issues occured. I fixed all of the ones that I saw, but ther might be more.


FrenchBen:
Thank you so much for putting this together, as it should solve some of the issues we're currently seeing.

Everything looks good to me. I've created a release with this PR, could you check that Ubuntu now works as expected?
[Kitematic Ubuntu](https://drive.google.com/file/d/1V8Eg20m35O5nFe3OuDHo9QCU7buTrZ4A/view?usp=sharing)

alex4o:
It looks fine to me.
kitematic
kitematic1

Issue #3223, opened 2017-10-27
the font is too small and con't copy log 


fairymaiden:
### Expected behavior
want easy to adjust font with the screen 
### Actual behavior
the font is too small and con't copy log 
### Steps to reproduce the behavior

  1. ...it is there 
  2. ...

collin5:
I would really love to work on this as my first PR

mhelmetag:
Go for it! It seems like their two issues here. Do you think you can tackle both?

collin5:
@mhelmetag Thanks :) 
I think I could be able to tackle both if that's ok

mhelmetag:
Heck yeah that's okay!

PR #3369, opened 2017-11-29
3223 Font too small and can't copy log.

collin5:
**What does this PR do?**

- Adds button for the user to copy container logs to the clipboard.
- Enables user to adjust the font size for logs.

[Fixes issue #3223 ]

screen shot 2017-11-29 at 12 01 13

### DIFF
--- a/src/components/ContainerHomeLogs.react.js
+++ b/src/components/ContainerHomeLogs.react.js
@@ -17,14 +17,14 @@ var FontSelect = React.createClass({
   getFontSizes: function(start, end){
     let options = [];
     for(let i = start; i<=end; i++){
-      options.push(<option value={i}>{i+" px"}</option>);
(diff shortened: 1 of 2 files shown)


collin5:
#3369

FrenchBen:
Thanks @collin5 for adding this - Could you remove the "box" around the buttons, so that they integrated better?

collin5:
@FrenchBen Thanks for your feedback, this should be fixed now 👍

FrenchBen:
Thanks @collin5 one more change before merge - Could we get the "Copied!" text to be white instead of Green? The green doesn't seem to fit with the overall theme

collin5:
Awesome 👍, let me do that ASAP

collin5:
@FrenchBen  Done 👍 . Thanks :)

PR #3441, opened 2017-12-23
[TypeScript] - Enable a TsConfig Ready

deinok:
This will allow use the type checkings of typescript.
Detectimng some common errors that JavaScript dont catch


package.json
```diff
@@ -11,16 +11,19 @@
...
+    "postinstall": "npm run build",
     "release": "grunt release",
     "release-verbose": "grunt --verbose release",
-    "lint": "eslint src"
+    "lint": "eslint src",
+    "tslint":"tslint --fix -p tsconfig.json"
   },
   "license": "Apache-2.0",
-  "electron-version": "0.35.4",
+  "electron-version": "1.7.10",
```
FrenchBen: Can you confirm that you  can still build and launch this on osx/windows? last update to Electron caused some issues for me.

deinok:
Seems working well.
TSC command added to all build pipelines of package.json . So any `npm` command will trigger the build
sin titulo

package.json
deinok: Yes, it is working well, but in not sure if this field is required now

FrenchBen:
How would this integrate with the current PR to have an actual windows installer generated?
https://github.com/docker/kitematic/pull/3447

deinok:
@FrenchBen This is completely independent of #3447 

This will enable us to check errors at compile time (Enabling the ability to use TypeScript and JSX for React).

deinok:
@FrenchBen Updated to work with Electron-Builder

FrenchBen:
thanks for the rebase!

deinok:
Perfect, next step is I will port about 1/2 of the code to TypeScript, in order to make the transition more natural.

FrenchBen:
Looks like I keep getting build errors when going through: 
`Running "shell:linux_npm" (shell) task`

Are you able to make a release from ubuntu?

deinok:
No, I can't, but it should not be related to this.

Seems some kind of error related only to the grunt targets.

This https://github.com/docker/kitematic/pull/3644 should allow anybody to generate the .zip files like when using grunt.

PR #3541, opened 2018-01-25
- Fixed favorites containers initialize.

danielamram:
Fix for the favorites containers feature.
Initialize the local history array with default an empty array.

Signed-off-by: Daniel Amram <danielamram0@gmail.com>


danielamram:
Ur welcome, let me know if there's more issues...

Issue #3625, opened 2018-03-05
Start on docker startup


N-Andronopoulos:
### Feature request

I think it's very essential to have a "start on docker startup" tick box, like doing to docker cli: `docker update --restart=always <CONTAINER ID>`

FrenchBen:
You mean having this option be part of the interface/ui?

N-Andronopoulos:
Yes of course, I think I would be best in the advanced settings of the container view (with the privileged, TTY stuff).

FrenchBen:
Feel free to play with it and provide us with a PR :)

FrenchBen:
@N-Andronopoulos Closing now, since the PR was merged

PR #3627, opened 2018-03-05
Added a tick box in advanced settings for restart policy to either 'a…

N-Andronopoulos:
#3625

### DIFF
--- a/src/components/ContainerSettingsAdvanced.react.js
+++ b/src/components/ContainerSettingsAdvanced.react.js
@@ -12,7 +12,7 @@ var ContainerSettingsAdvanced = React.createClass({
   },
 
   getInitialState: function () {
-    let [tty, openStdin, privileged, restartPolicy] = ContainerUtil.mode(this.props.container) || [true, true, false, {MaximumRetryCount: 0, Name: ''}];
+    let [tty, openStdin, privileged, restartPolicy] = ContainerUtil.mode(this.props.container) || [true, true, false, {MaximumRetryCount: 0, Name: 'no'}];
     return {
       tty: tty,
       openStdin: openStdin,
@@ -26,7 +26,7 @@ var ContainerSettingsAdvanced = React.createClass({
     let tty = this.state.tty;
     let openStdin = this.state.openStdin;
     let privileged = this.state.privileged;
-    let restartPolicy = this.state.restartPolicy? {MaximumRetryCount: 0, Name: 'always'} : {MaximumRetryCount: 0, Name: ''};
+    let restartPolicy = this.state.restartPolicy? {MaximumRetryCount: 0, Name: 'always'} : {MaximumRetryCount: 0, Name: 'no'};
     let hostConfig = _.extend(this.props.container.HostConfig, {Privileged: privileged, RestartPolicy: restartPolicy});
     containerActions.update(this.props.container.Name, {Tty: tty, OpenStdin: openStdin, HostConfig: hostConfig});
   },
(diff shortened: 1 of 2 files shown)


src/components/ContainerSettingsAdvanced.react.js
```diff
@@ -60,6 +68,7 @@ var ContainerSettingsAdvanced = React.createClass({
             <p><label><input type="checkbox" checked={this.state.tty} onChange={this.handleChangeTty}/>Allocate a TTY for this container</label></p>
             <p><label><input type="checkbox" checked={this.state.openStdin} onChange={this.handleChangeOpenStdin}/>Keep STDIN open even if not attached</label></p>
             <p><label><input type="checkbox" checked={this.state.privileged} onChange={this.handleChangePrivileged}/>Privileged mode</label></p>
+            <p><label><input type="checkbox" checked={this.state.restartPolicy} onChange={this.handleChangeRestartPolicy}/>Start on docker startup</label></p>
```
FrenchBen: I would perhaps label this differently, something like `Enable 'always' restart policy`
By definition, the `always` policy does a lot more than just start when docker starts:
```
Always restart the container regardless of the exit status. When you specify always, the Docker daemon will try to restart the container indefinitely. The container will also always start on daemon startup, regardless of the current state of the container.
```

src/components/ContainerSettingsAdvanced.react.js
```diff
@@ -25,7 +26,8 @@ var ContainerSettingsAdvanced = React.createClass({
     let tty = this.state.tty;
     let openStdin = this.state.openStdin;
     let privileged = this.state.privileged;
-    let hostConfig = _.extend(this.props.container.HostConfig, {Privileged: privileged});
+    let restartPolicy = this.state.restartPolicy? {MaximumRetryCount: 0, Name: 'always'} : {MaximumRetryCount: 0, Name: ''};
```
FrenchBen: default should be `Name: "no"` not `Name: ''`

src/utils/ContainerUtil.js
```diff
@@ -18,7 +18,8 @@ var ContainerUtil = {
     return [
         (container && container.Config) ? container.Config.Tty : true,
         (container && container.Config) ? container.Config.OpenStdin : true,
-        (container && container.HostConfig) ? container.HostConfig.Privileged : false
+        (container && container.HostConfig) ? container.HostConfig.Privileged : false,
+        (container && container.HostConfig) ? container.HostConfig.RestartPolicy : {MaximumRetryCount: 0, Name: ''}
```
FrenchBen: Default should be `Name: "no"`, not `Name: ''`

N-Andronopoulos:
I agree about the tick label, I shall change it right away. But...

Cording to _https://docs.docker.com/engine/api/v1.24/_
> RestartPolicy – The behavior to apply when the container exits. The value is an object with a Name property of either "always" to always restart, "unless-stopped" to restart always except when user has manually stopped the container or "on-failure" to restart only when the container exit code is non-zero. If on-failure is used, MaximumRetryCount controls the number of times to retry before giving up. The default is not to restart. (optional) An ever increasing delay (double the previous delay, starting at 100mS) is added before each restart to prevent flooding the server.

There is no 'no' policy, that's why I left it blank. Also its the default value after you create a new container in docker, by doing a docker inspect.

Also the 'always' policy doesn't stop you from stopping the container, but as you said after a restart it will try indefinitely.

FrenchBen:
Interesting - I was looking at the CLI doc, which provides 4 options:
https://docs.docker.com/engine/reference/run/#restart-policies-restart

I also did a quick inspect on a container I had running, without any `restart` policy attached; the JSON data shows:
```
"RestartPolicy": {
                "Name": "no",
                "MaximumRetryCount": 0
}
```
Looks like both are accepted:
https://github.com/moby/moby/blob/master/api/types/container/host_config.go#L262

In which case, I prefer to see `"no"` being passed, as it seems more explicit.

N-Andronopoulos:
Ok, done :)

PR #3644, opened 2018-03-10
Electron-Builder - Windows & Linux & Mac - ZIP

deinok:
.zip files are added as distribute targets to all SOs.

Resulting files have been tested in Windows 10 and in Ubuntu 16.04


FrenchBen:
While trying to run the above to release a zip, I get the following error:
```
>> Local Npm module "grunt-electron-installer-debian" not found. Is it installed?
>> Local Npm module "grunt-electron-installer-redhat" not found. Is it installed?
Warning: Task "electron-installer-debian:linux64" not found. Use --force to continue.
```

deinok:
Yes, this .zip are generated using the electron-builder.
So use "npm run release:{yourPlataform}" to generate the .zip and other installers

FrenchBen:
@deinok Usually though, all archives are made from 1 OS (Mac OSX), which currently doesn't seem possible?

deinok:
Well, yes, there are some issues with this. This are the issues:
In order to build packages in the format of: msi, deb, pac, etc you have to build it in the SO that you are targeting.

MacOS can only build on MacOS
Linux is builded on any plataform (Remote compilation)
AppX only on Win10. Other Targets can be build on any Win.

To build for linux you exec "npm run release:linux" and for win you run "npm run release:windows", etc...

The best way is to use CIs to make the releases.
There is no way to fix this due SO issues, not related to kitematic, electron or electron-builder

FrenchBen:
@deinok We were able to do this previously without any issues. I didn't think that our conversion to TS would imply these limitations?

deinok:
Well, TS things are not related to electron-builder.
I can try to support the old release way, but the old way would not be compat with creating app packages

FrenchBen:
@deinok any chance we can leverage containers to do this? Aka we boot up a container that has all of the packages needed to make this work?

deinok:
Well, there are a electron-builder images for building linux and win (not appx).
But i dont think there are any image for macOS.

I can try sto search some other solutions

FrenchBen:
That'd be great! Thanks a lot for your help

deinok:
Ok, I have added a package.script called "release:all-zip".

This should build only .zip for all plataforms (But it will only completely in MacOS due MacOS requirement).

Here is more information about building for all plataforms: https://www.electron.build/multi-platform-build
Seems like we are going to need 3 diferent CIs. :(

Also, I'm going to open an issue in electron-builder to try to avoid the MacOS requirements with only .zip packages

PR #3950, opened 2018-07-04
Fix tslint path in package.json

samtux:
Fix tslint path. Issue in https://github.com/palantir/tslint/issues/1664


jeanlaurent:
Hi @samtux can you confirm this work with the fixed tslint version of master ?

PR #4063, opened 2018-08-23
4009 Fix can not create container from private image

### LINKED ISSUE #4009: Kitematic does not read full image name
We have some private docker repos set up with Artifactory and the naming convention we use is organization/image-name:tag

The full image image ends up looking like this:
repo-url.net/org/image:tag

### Expected behavior
Kitematic sees the image name

### Actual behavior
Kitematic displays the url and org. The image name isn't displayed and isn't used when I click create on an image.
### Information about the Issue
I am able to start a container for the image using the command line.

  1. Create an image using repo-url/something/something-else:tag
  2. Try to create a container with the image using the New button and My Images

ssfyn:
fix #4009

### DIFF
--- a/src/utils/DockerUtil.js
+++ b/src/utils/DockerUtil.js
@@ -298,7 +298,9 @@ var DockerUtil = {
           if (repo.indexOf('/') === -1) {
             repo = 'local/' + repo;
           }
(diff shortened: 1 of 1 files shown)


tscswcn:
something useful

PR #4989, opened 2019-06-21
Fix v0.17.7 container log error

ssfyn:
Kitematic v0.17.7 upgrade "dockerode" module to 2.5.8,  Brought a new problem: when opts.follow is false, result type changed from IncomingMessage to Buffer @jdrouet   
my os is MacOS


guillaumerose:
Thanks for your PR!

PR #5325, opened 2019-09-30
bump dockerode version to 3.0.1 & add default user-agent  to dockerod…

glours:
…e http requests

Signed-off-by: Guillaume Lours <guillaume.lours@docker.com>

### DIFF
--- a/src/utils/DockerUtil.js
+++ b/src/utils/DockerUtil.js
@@ -48,7 +48,7 @@ var DockerUtil = {
 
     if (ip.indexOf('local') !== -1) {
       try {
-        this.client = new dockerode({socketPath: this.socketPath});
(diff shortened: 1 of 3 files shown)


src/utils/DockerUtil.js
```diff
@@ -64,7 +64,8 @@ var DockerUtil = {
         port: 2376,
         ca: fs.readFileSync(path.join(certDir, 'ca.pem')),
         cert: fs.readFileSync(path.join(certDir, 'cert.pem')),
-        key: fs.readFileSync(path.join(certDir, 'key.pem'))
+        key: fs.readFileSync(path.join(certDir, 'key.pem')),
+        headers: {'user-agent': 'docker-desktop'},
```
guillaumerose: docker-desktop? it should be kitematic no?

glours: definitely, just push the correction

PR #5611, opened 2020-04-03
Container logs font size can be saved to the localStorage

microleft:
Kitematic is really great tool, but the default font size(10px) is very uncomfortably small for me
(Windows 10 2560x1440 resolution ),this **PR** do: 
- Enables user to **save** the font size


rmcsharry:
I cannot even read the logs at my age, so they are useless to me. +1

rmcsharry:
I think, though, that this is a duplicate issue

https://github.com/docker/kitematic/issues/1316

